PR #65, opened 2016-09-09
Add automatic deployment

fspoettel:
- Release to npm on new tag via Travis integration
- Release new version of docs on push to master
- Fix asset-paths in docs-app

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "description": "Salesforce Lightning Design System React Components",
   "keywords": [
     "react",
-    "react-component",
     "salesforce lightning design system",
     "lightning design system",
     "slds",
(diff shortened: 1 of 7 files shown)


wwwdata:
I currently don't understand how the es5 transpiling is triggered before `npm publish` which is made with the travis plugin. Did you forget to do that?

wwwdata:
ah, and don't comment with the ci user 😄

fspoettel:
sorry about that! 😓 you are right, the whole transpiling step should be set as prepublish-script. i am surprised that this is not the case anyways. I will add it and update the documentation accordingly.

src/index.js
```diff
@@ -5,7 +5,7 @@ export * from './components/Button';
 export * from './components/ButtonGroup';
 export * from './components/Card';
 export * from './components/DataTable';
-export * from './components/Form';
+  export * from './components/Form';
```
wwwdata: whät?

package.json
```diff
@@ -15,12 +14,15 @@
...
     "build": "npm run build:babel && npm run build:copy-files",
     "build:babel": "babel ./src --ignore *.spec.js --out-dir ./dist",
     "build:copy-files": "cp ./src/styles.css ./dist/; cp ./README.md ./dist/",
-    "build:docs": "webpack --config ./docs/webpack.prod.config",
+    "build:docs": "rm -rf ./docs/build; webpack --config ./docs/webpack.prod.config",
+    "deploy": "bash deploy.sh",
     "prebuild": "rm -rf ./dist",
-    "lint": "eslint src docs"
+    "prepublish": "npm run prebuild; npm run build",
+    "pretest": "npm run lint"
```
wwwdata: this is actually awesome. Force ppl to use damn linter. Nobody runs it locally I guess

src/index.js
fspoettel: i broke it on purpose to check if pretest correctly returns the error-code for travis

fspoettel:
I tested the gh-pages deployment locally without the travis-stuff like this: https://gist.github.com/fspoettel/59b641f3acbd24748c06ce624f54d416 and it works. Not sure if i got the travis- tuff right tho - only a push to master can tell I guess.

PR #66, opened 2016-09-09
Fix Travis Token

fspoettel:
This works when `GITHUB_TOKEN` is set via the web interface for Travis, but I can't for the life of me figure out how to add an encrypted token for this specific repository. These are the steps I've taken:

- Get an access-token for `propertybase-ci` (this token works when set in Web UI)
- Install the travis gem, log into propertybase-ci by running `travis login --pro`
- Encrypt the token by running `travis encrypt "GITHUB_TOKEN={token}"` I also tried explicitely setting the repo by passing `-r propertybase/react-lds`, does not make a difference.
- Add the token to `env.global`

If someone could point me in the right direction on this, it would be awesome. Maybe this is related to the repo being public and hosted on `travis-ci.com` instead of `travis-ci.org`. I tried to test it with `travis login` (to get a token for the public CI) but `propertybase-ci` needs to authorize with `travis-ci.org` first and I want to confirm that this is ok before giving the access.

**Edit**: I implemented a similiar configuration with a personal open-source project hosted on `travis-ci.org` and it works without issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,11 @@ language: node_js
 cache:
   bundler: true
   directories:
-    - node_modules
+  - node_modules
 node_js:
-  - "6"
-env:
-  global:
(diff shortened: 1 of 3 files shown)


fspoettel:
@konsti this already works and I would consider it done in the sense that the functionality is all there. When the repository is moved to public travis (see above), I'll add the encrypted key in the travis.yml instead of relying on the web interface for that.

chriserik:
👍

PR #68, opened 2016-09-26
Custom Button in Email Footer

wwwdata:
Allow usage of a completely custom button in Email footerButtons

### DIFF
--- a/docs/src/app/pages/DockedComposer/Email/ExampleEmail.js
+++ b/docs/src/app/pages/DockedComposer/Email/ExampleEmail.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Email } from 'react-lds';
+import { Email, Button, ButtonIcon } from 'react-lds';
 
 const footerButtons = [
   {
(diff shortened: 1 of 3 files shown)


docs/src/app/pages/DockedComposer/Email/ExampleEmail.js
```diff
@@ -24,6 +24,11 @@ const footerButtons = [
     icon: 'delete',
     key: 'delete_icon',
   },
+  {
+    customButton: <Button icon icon-container key="CustomButton">
```
fspoettel: It may be better to set the key automatically by using `React.cloneElement()`, but I'm not sure whether there is a significant performance penalty when using that fn

wwwdata: ok, it's just in the docs anyway

fspoettel: i meant setting it in the loop automatically for each customButton, but having to set it is fine since it is documented.

PR #70, opened 2016-09-27
Advanced Modal Table Layout for Lookup

wwwdata:
For the totally unstable table lookup layout here: https://lightningdesignsystem.com/components/lookups/#flavor-advanced-modal

but we need this inside a modal box, otherwise the modal box stays very small und the user has to scroll with the normal Lookup.

### DIFF
--- a/src/components/Lookup/Lookup.js
+++ b/src/components/Lookup/Lookup.js
@@ -13,6 +13,9 @@ import {
   InputRaw,
   Pill,
   PillContainer,
+  Table,
+  Row,
+  Cell,
 } from '../../';
 
 const validateSelection = (props, propName, componentName, ...rest) => {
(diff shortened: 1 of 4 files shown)


src/components/Lookup/Lookup.js
```diff
@@ -415,26 +502,22 @@ export class Lookup extends React.Component {
...
+          {this.selections()}
+          {this.lookupList()}
+        </FormElement>
+        {this.lookupListTable()}
+      </div>
     );
   }
 }
 
-Lookup.defaultProps = {
```
fspoettel: nice catch :)

src/components/Lookup/Lookup.js
```diff
@@ -93,6 +96,28 @@ export class Lookup extends React.Component {
      * timestamp.
      */
     allowCreate: React.PropTypes.bool,
+    /**
```
fspoettel: We should be careful of adding experimental options to otherwise stable components but personally I'm ok with it. We probably have to revisit this once SF moved the Lookup-Modal to stable

wwwdata: yes.. but the thing is that we need this now so I have to do it, sorry ❤️

PR #78, opened 2016-10-11
Fix docs

fspoettel:
- Use resolver to match multiple exported components
- Update ESLint


package.json
```diff
@@ -67,10 +67,9 @@
     "enzyme": "^2.4.1",
     "eslint": "^3.7.1",
     "eslint-config-airbnb": "^12.0.0",
-    "eslint-import-resolver-webpack": "~0.5.1",
```
wwwdata: hm? this is not needed anymore so that eslint can check imports?

fspoettel: 🙈fixed

PR #79, opened 2016-10-12
Navigation Component

wwwdata:
also updates slds to `2.1.3`. The Navigation was introduced in `2.1.2`

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "scripts": {
     "start": "node ./docs/server.js",
     "test": "jest",
-    "test-watch": "jest --watch=all",
+    "test-watch": "jest --watchAll",
     "lint": "eslint src docs",
     "build": "npm run build:babel && npm run build:copy-files",
     "build:babel": "babel ./src --ignore *.spec.js --out-dir ./dist",
@@ -52,7 +52,7 @@
     "react-click-outside": "2.1.0"
   },
   "devDependencies": {
-    "@salesforce-ux/design-system": "~2.0.3",
+    "@salesforce-ux/design-system": "2.1.3",
     "babel-cli": "^6.14.0",
     "babel-core": "^6.14.0",
     "babel-eslint": "^6.1.2",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -52,7 +52,7 @@
     "react-click-outside": "2.1.0"
   },
   "devDependencies": {
-    "@salesforce-ux/design-system": "~2.0.3",
+    "@salesforce-ux/design-system": "2.1.3",
```
fspoettel: it's probably a good idea to lock down slds further 👍

src/components/Navigation/NavigationHeadline.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
fspoettel: why is this a standalone component? we use a `title` or `label`-prop for header's in similiar components.

src/components/Navigation/NavigationHeadline.js
```diff
@@ -0,0 +1,31 @@
...
+
+NavigationHeadline.propTypes = {
+  /**
+   * id of the headline element
+   */
+  id: React.PropTypes.string,
+  /**
+   * headline value
+   */
+  value: React.PropTypes.string.isRequired,
```
fspoettel: if there is a good reason to split this into it's own module, it would be nice to rename this prop to `title` to be consistent with the rest of the components

src/components/Navigation/index.js
```diff
@@ -0,0 +1,3 @@
+export { default as Navigation } from './Navigation';
```
fspoettel: you only need to do this if there is a named export in the form of `export const Component`. Otherwise, `export Component from './'` is sufficient.

wwwdata: yep, but for the main `Navigation` element I got to do this because I export a variationable version and a normal one. Will change for others

wwwdata:
@fspoettel pushed again. Now there is a `NavigationList` component that is the parent for `NavigationListElement`

PR #80, opened 2016-10-12
LDS 2.1.3

fspoettel:
- Menus are incosistent with the ARIA-standards in the new SLDS as it would need breaking changes to update it


src/components/Form/__tests__/Input.spec.js
```diff
@@ -112,6 +112,11 @@ describe('<Input />', () => {
     expect(mounted.find('div.slds-form-element__control IconSVG').first().props().icon).toEqual('warning');
   });
 
+  it('labels the input with an error', () => {
+    mounted.setProps({ error: 'shit' });
```
wwwdata: :trollface:

PR #84, opened 2016-10-27
Switch to yarn

wwwdata:
There's currently one interesting discussion ongoing: https://github.com/yarnpkg/yarn/issues/570

the behaviour at the moment is that when a normal `yarn install` is running, it automatically generates and updates the `yarn.lock` file within the specified version range from the `package.json`

So in order to reproduce the correct dependencies always purely from the lock file it's important to add `--pure-lockfile` to the `yarn install` command ... I hope they change that. But just in case people are wondering here why the hell the yarn.lock changes after an install.


package.json
```diff
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "classnames": "^2.2.5",
+    "eslint": "3.7.1",
```
wwwdata: @fspoettel don't your just thumbs up, give me a real review and approve!!! 😄

fspoettel: i thought i did, looks like i did not push the Approve button hard enough

wwwdata: it's ok, just make sure you are fully concentrated next time. It's important to stay focused when pressing buttons!

fspoettel: giphy

PR #85, opened 2016-10-28
Fixes wrong variable name in Readme

chriserik:


wwwdata:
tests are failing because the `eslint` guys fucked up a rule. Let's merge my #84  stuff and then you rebase this on it and it will work

PR #86, opened 2016-11-04
Add callback on lookup resultAdd

fspoettel:
Adds a callback that is executed when a result is added in `Lookup`.

The callback is passed the result pill and according item from the state and applies a cloned node if the callback returns one.

This is needed for PHOENIX-447 and I do not see another way to implement this besides completely rewriting the component to be more modular (something we should probably do anyways sooner or later)

@wwwdata I'd like to have your input on whether this is too much of a hack when you are back to health.

Also done in this PR:

 - Cleaned up callbacks in Lookup, removed all binds in `render`
 - Shortened the debounce delay in `Lookup` to 400ms
 - Removed a duplicate dependency for eslint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Next
+* Add callback when result is added to Lookup. It can manipulate the result that is added
+* Clean up callback bindings in Lookup & reduce debounce delay by 100ms
+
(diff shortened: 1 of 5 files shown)


wwwdata:
we don't do this color thing anymore. So if you think this is shit (it's not specified for lightning in any way) we can also drop this PR

PR #89, opened 2016-11-23
Update datepicker

PeterSiegl:


src/components/Datepicker/Datepicker.js
```diff
@@ -124,6 +124,16 @@ export class Datepicker extends React.Component {
     return (
       <tbody>
         {this.getMonthDays().map((week, weekIndex) => this.renderWeek(week, weekIndex))}
+        <tr>
```
wwwdata: this was outside the table body, good catch 😂

wwwdata: not that my review counts anything, you still cannot merge!

PR #91, opened 2016-12-12
Create elements leaving focus of lookup component

PeterSiegl:


src/components/Lookup/Lookup.js
```diff
@@ -197,21 +198,50 @@ export class Lookup extends React.Component {
...
+    this.setState({ selected, searchTerm: '', open: false });
+  }
+
+  handleCreateElementOnBlur(e) {
+    // if no result was found, allow creation of new element on leave
+    const value = e.target.value;
+    if (this.props.emailLayout) {
+      // basic input validation in case of email layout
+      if (
+        /^.+@.+\..+$/.test(value) &&
```
fspoettel: not a big fan of doing this here, it's the wrong place imo. this should be moved once we refactor quicksend. also, we have to look into a form component with validation for the react-lds, maybe we can centralize it for all of our UI components.

PR #93, opened 2017-01-25
Add popovers

PeterSiegl:


src/components/Popover/Popover.js
```diff
@@ -0,0 +1,168 @@
...
+    customLayout,
+    ...rest,
+  } = props;
+  const prefix = (classes, passThrough) => prefixClasses(cssPrefix, classes, passThrough);
+
+  const sldsClasses = [
+    'popover',
+    { 'nubbin--left': !nubbin },
+    { [`nubbin--${nubbin}`]: !!nubbin },
+    { 'popover--panel': (!!panels && (typeof customLayout === 'undefined')) },
```
fspoettel: props are nullable. it would make sense to incorporate that into the check

PR #96, opened 2017-02-15
Add outside click support

PeterSiegl:


fspoettel:
@konsti a class is needed here because react-click-outside needs a ref and stateless functions do not support those

PeterSiegl:
@konsti I moved the propTypes outside. can I merge this one?

PR #101, opened 2017-03-13
Fix font styles

PeterSiegl:


src/components/Popover/Popover.js
```diff
@@ -36,7 +36,7 @@ export class Popover extends React.Component {
     let headerContent;
     let borderRadius;
     if (!!panels && (typeof customLayout !== 'undefined') && customLayout !== '') {
-      headerContent = (<h2 className={this.prefix(['text-heading--medium', 'p-around--x-small'])}>{header}</h2>);
+      headerContent = (<h2 className={this.prefix(['text-heading--small p-around--x-small'])}>{header}</h2>);
```
fspoettel: missing array comma here

PeterSiegl: good catch!

PR #104, opened 2017-03-20
Make `slds-col` optional for Columns & remove deprecated flavors, closes #103

### LINKED ISSUE #103: `Column` does not support omitting `slds-col`
See: https://github.com/propertybase/react-lds/blob/master/src/components/Grid/Column.js#L24
See: https://lightningdesignsystem.com/components/utilities/grid/#flavor-automatic-sizing-no-stretch

- [ ] Column should set slds-col by default, but have the option to not set it
- [ ] Remove deprecated flavors, see: https://lightningdesignsystem.com/components/utilities/grid/#overview
- [ ] Check if there are new flavors
- [ ] Update `pba-react` components that use deprecated flavors or could benefit from the improved `Column`

tomlord90:


src/components/Grid/Column.js
```diff
@@ -17,13 +17,14 @@ const breakPointProp = breakpoint => `${breakpoint}-sizeOf`;
 const sizeRegex = /^([1-9]|1[0-2])-([1-9]|1[0-2])$/;
 
 export const Column = (props, { cssPrefix }) => {
-  const { align, children, className, ...rest } = props;
+  const { align, children, className, omitSldsCol, ...rest } = props;
```
fspoettel: `omitCol` is sufficient here

src/components/Grid/Column.js
```diff
@@ -17,13 +17,14 @@ const breakPointProp = breakpoint => `${breakpoint}-sizeOf`;
...
+
   const prefix = (classes, passThrough) => prefixClasses(cssPrefix, classes, passThrough);
 
-  const sldsClasses = [
-    'col',
-    { [`align-${align}`]: !!align },
-  ];
+  const sldsClasses = [];
```
fspoettel:
``` js
const sldsClasses = [
  { col: !omitCol },
  { [`align-${align}`]: !!align }
];
```
We are using the `classNames` library here, in order to not having to manipulate the class array with array methods. See above code for an example of usage.

fspoettel:
Nice, this is almost good-to-go. Please write a unit test for the behavior. Check the other specs or the enzyme/jest documentation for information on how to do this.

PeterSiegl:
Please squash commits then 👍

PR #112, opened 2017-04-11
Update to react 15.5

fspoettel:
This updates the propTypes access. Some considerations:

- Enzyme is not fully updated yet, which means a couple of console.errors will be thrown while testing. This should be fixed with the next version
- We should get rid of react-click-outside as we want to use those components in a controlled fashion and it still relies on React.createClass. The dropdown refactor would be a good time for that


fspoettel:
Danke Ben! ;) Allet jut bei dir?  @wwwdata

PR #115, opened 2017-05-02
[NEW] Propertybase/release/1.1.6

carlo:
This PR adds the `hideLabel` prop to components `Checkbox`, `CheckboxRaw`, `Select`, `Textarea`.


src/components/Form/CheckboxRaw.js
```diff
@@ -31,10 +32,14 @@ const CheckboxRaw = (props, { cssPrefix }) => {
     />
 );
 
+  const labelClassNames = ['form-element__label'];
```
fspoettel: you could also use a classNames array here

PR #125, opened 2017-05-17
Add localization and timezone, update props and validation

PeterSiegl:


src/components/Datepicker/Datepicker.js
```diff
@@ -31,97 +32,137 @@ export class Datepicker extends React.Component {
...
 
-  /**
-   * Callback when the month changes
-   * @param  {Number} relativeMonth one month forward or backward
-   * @return {void}                 interacts with component state
-   */
-  onMonthChange = (relativeMonth) => {
-    const { viewedDate } = this.state;
+    const output = isValid ? nextDate.format(iso8601DateFormat) : null;
+    onChange(output, isValid);
```
carlo: This will fire every time a prop is changed, whether it was an actual date change or not.  I'd specifically add a string comparison between `nextProps.date` and the current `date` to see if the `onChange()` call is necessary.

src/components/Datepicker/Datepicker.js
```diff
@@ -31,16 +32,120 @@ export class Datepicker extends React.Component {
...
+
+    let initial;
+    let isValid;
+
+    if (date) {
+      initial = date;
+      isValid = moment(date, iso8601DateFormat, true).isValid();
+    } else if (!date && initialDate) {
+      initial = initialDate;
+      isValid = moment(initialDate, iso8601DateFormat, true).isValid();
```
fspoettel: This seems to be used a lot, maybe make it a static method?

``` js
static validate = (date) => moment(date, iso8601DateFormat, true).isValid()
```

src/components/Datepicker/Datepicker.js
```diff
@@ -31,16 +32,120 @@ export class Datepicker extends React.Component {
...
       open,
-      preselectedDate: preselectedDate ? moment(preselectedDate) : null,
-      inputDate: preselectedDate ? moment(preselectedDate).format(salesforceDateFormat).toString() : '',
-      viewedDate: moment(),
+      inputValue: isValid ? moment(initial).format(defaultDateFormat).toString() : '',
+      isInvalid: !isValid,
+      viewedDate: isValid ? moment(initial) : moment(),
+    };
+
+    this.mergedTranslations = {
```
fspoettel: Relying on defaultProps here seems dangerous. Abstract the default props into a constant and merge that

src/components/Datepicker/Datepicker.js
```diff
@@ -31,16 +32,120 @@ export class Datepicker extends React.Component {
...
 
+  componentWillReceiveProps(nextProps) {
+    const { date } = nextProps;
+    const nextDate = moment(date, [iso8601DateFormat, defaultDateFormat]);
+    const isValid = moment(date, iso8601DateFormat, true).isValid();
+
+    this.setState({
+      inputValue: isValid ? nextDate.format(defaultDateFormat).toString() : date,
+      isInvalid: !isValid,
+      isInputInvalid: !isValid,
```
fspoettel: Why are there two state keys for this?

PR #128, opened 2017-05-24
Fix highlighted days, consider required state, fix focus event

PeterSiegl:


src/components/Datepicker/Datepicker.js
```diff
@@ -34,8 +38,27 @@ export class Datepicker extends React.Component {
...
+   * @return {Bool}
+   */
+  static isControlled = date => typeof date !== 'undefined';
+
+  /**
+   * Checks whether the component is controlled
+   * @param {String} date date as string or moment object
+   * @return {Bool}
+   */
+  static isControlled = date => typeof date !== 'undefined';
```
fspoettel: why is there a dupe here?

src/components/Datepicker/Datepicker.js
```diff
@@ -57,19 +80,20 @@ export class Datepicker extends React.Component {
...
+    } else {
       initial = initialDate;
       isValid = Datepicker.validate(initialDate);
     }
 
     this.state = {
       open,
       inputValue: isValid ? moment(initial).format(defaultDateFormat).toString() : '',
-      isInvalid: !isValid,
+      isValid: true,
```
fspoettel: Why is this initialized as true if isValid is checked above? I don't understand

src/components/Datepicker/Datepicker.js
```diff
@@ -80,16 +104,21 @@ export class Datepicker extends React.Component {
...
-    });
+    if (isControlled) {
+      const nextValidDate = moment(nextDate, [iso8601DateFormat, defaultDateFormat]);
+      const isValid = Datepicker.validate(nextDate);
+
+      this.setState({
+        inputValue: isValid ? nextValidDate.format(defaultDateFormat).toString() : nextDate,
+        isValid: nextDate === null || nextDate === '' ? !required : isValid,
+        viewedDate: isValid ? nextValidDate : moment(),
+        open: nextDate === null || nextDate === '' ? !!required : !isValid,
```
fspoettel: Same condition as isValid, use a const

src/components/Datepicker/Datepicker.js
PeterSiegl: because initially we don't want to see an error message. but as soon as the user clicks on a required input field or enters a wrong date, the error message gets displayed.

PR #158, opened 2017-06-21
Add expandable section new

PeterSiegl:
Related: #146


src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  renderCollapsable() {
+    const { id, title } = this.props;
+    const { open } = this.state;
+
+    return (
+      <Button
+        aria-controls={id}
+        aria-expanded={open.toString()}
```
fspoettel: Use: `${open}`

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,134 @@
...
+    ];
+
+    return (
+      <div className={cx(sldsClasses)}>
+        <h3 className={cx(headerClasses)}>
+          {collapsable
+            ? this.renderCollapsable()
+            : this.renderUncollapsable()}
+        </h3>
+        <div aria-hidden={(!open).toString()} className="slds-section__content" id={id}>
```
fspoettel: Use `${!open`

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,134 @@
...
+          {children}
+        </div>
+      </div>
+    );
+  }
+}
+
+ExpandableSection.defaultProps = {
+  className: null,
+  title: null,
```
fspoettel: Should be required and no defaultProp

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,134 @@
...
+   */
+  open: PropTypes.bool,
+  /**
+   * triggered when the user clicks the expand button
+   */
+  onClickToggle: PropTypes.func,
+  /**
+   * content to be displayed
+   */
+  children: PropTypes.node,
```
fspoettel: Should be required

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,134 @@
...
+        'When collapsable is false, open cannot be false.',
+      );
+    }
+
+    this.state = {
+      open,
+    };
+  }
+
+  onClickToggle = () => {
```
fspoettel: Call prop `onClickToggle` here too.

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,134 @@
...
+
+    if (!open && !collapsable) {
+      // eslint-disable-next-line
+      console.warn(
+        '[react-lds] ExpandableSection:',
+        'When collapsable is false, open cannot be false.',
+      );
+    }
+
+    this.state = {
```
fspoettel: This component should be usable as a `Controlled` and an `Uncontrolled` component. If a component is uncontrolled, it manages its own state (what you are doing right now). The problem here is that there is no interface to open and close the section programatically. The way to fix this is as follow:

 - If `defaultOpen` is passed, apply it to `this.state` and and let the component handle it from there
 - If `open` is passed, use the `open` prop instead of the internal state and simply render it. 
 - If there is a conflict with collapsible, overwrite `defaultOpen` in uncontrolled mode and add a warning in controlled mode

fspoettel:
@simodrws status on this?

simodrws:
@fspoettel work in progress, will upload later

simodrws:
I am a little unsure about the tests, some of them fail because of the **defaultOpen** and **open** property being null. It would be nice if someone could help me with that.

src/components/ExpandableSection/__tests__/ExpandableSection.spec.js
```diff
@@ -0,0 +1,53 @@
...
+
+import { Button } from '../../../';
+import ExpandableSection from '../ExpandableSection';
+
+const getExpandableSection = (props = {}) => shallow(<ExpandableSection id="51er" {...props} />);
+
+
+describe('<ExpandableSection />', () => {
+  it('renders right class name', () => {
+    const mounted = getExpandableSection();
```
fspoettel: `const mounted = getExpandableSection({ open: true });` no need to use setProps here if you use a fn to get the component

src/components/ExpandableSection/__tests__/ExpandableSection.spec.js
```diff
@@ -0,0 +1,53 @@
...
+    mounted.setProps({ open: true });
+    expect(mounted.hasClass('slds-section')).toBeTruthy();
+  });
+
+  it('renders children', () => {
+    const child = <div className="51ers" />;
+    const mounted = getExpandableSection({ id: 'test', children: child });
+    mounted.setProps({ open: true });
```
fspoettel: You don't supply `open` or `defaultOpen` when mounting the component and hence get a warning in the console, you should supply one of them at mount if your component needs it to function

src/components/ExpandableSection/__tests__/ExpandableSection.spec.js
```diff
@@ -0,0 +1,53 @@
...
+    expect(mounted.find('.slds-truncate').hasClass('slds-p-horizontal_small')).toBeTruthy();
+  });
+  it('renders right slds-classes when defaultOpen is set to true', () => {
+    const mounted = getExpandableSection();
+    mounted.setProps({ defaultOpen: true });
+    expect(mounted.find('.slds-section').hasClass('slds-is-open')).toBeTruthy();
+  });
+  it('renders right slds-classes when defaultOpen is set to null', () => {
+    const mounted = getExpandableSection();
+    mounted.setProps({ defaultOpen: null, open: true });
```
fspoettel: `defaultOpen` is `null` by default

src/components/ExpandableSection/__tests__/ExpandableSection.spec.js
```diff
@@ -0,0 +1,53 @@
...
+    const mounted = getExpandableSection({ open: false });
+    expect(mounted.find('.slds-section').hasClass('slds-is-open')).toBeFalsy();
+  });
+  it('does render right class when collapsable is false', () => {
+    const mounted = getExpandableSection({ collapsable: false });
+    expect(mounted.find('.slds-truncate').hasClass('slds-p-horizontal_small')).toBeTruthy();
+  });
+  it('renders right slds-classes when defaultOpen is set to true', () => {
+    const mounted = getExpandableSection();
+    mounted.setProps({ defaultOpen: true });
```
fspoettel: `const mounted = getExpandableSection({ defaultProps: true });` no need to use setProps here if you use a fn to get the component

src/components/ExpandableSection/__tests__/ExpandableSection.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  it('renders children', () => {
+    const child = <div className="51ers" />;
+    const mounted = getExpandableSection({ id: 'test', children: child });
+    mounted.setProps({ open: true });
+    expect(mounted.contains(child)).toBeTruthy();
+  });
+
+  it('renders a title', () => {
+    const mounted = getExpandableSection();
+    mounted.setProps({ title: 'Title', open: false });
```
fspoettel: `const mounted = getExpandableSection({ title: 'Title', open: false });` no need to use setProps here if you use a fn to get the component

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,145 @@
...
+
+import { Button, ButtonIcon } from '../../';
+
+class ExpandableSection extends Component {
+  constructor(props) {
+    super(props);
+    const { open, collapsable } = props;
+
+    if (props.defaultOpen != null) {
+      if (!collapsable) this.state = { open: true };
```
fspoettel: Please use braces

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,145 @@
...
+import cx from 'classnames';
+
+import { Button, ButtonIcon } from '../../';
+
+class ExpandableSection extends Component {
+  constructor(props) {
+    super(props);
+    const { open, collapsable } = props;
+
+    if (props.defaultOpen != null) {
```
fspoettel: `!==`

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,145 @@
...
+      if (!collapsable) this.state = { open: true };
+      const defaultOpen = this.props.defaultOpen;
```
fspoettel: see above, not needed

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,145 @@
...
+      this.state = { open: defaultOpen };
+    }
+
+    if (props.defaultOpen == null) {
```
fspoettel: `===`

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,145 @@
...
+      if (open != null) {
```
fspoettel: `!==`

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,145 @@
...
+        console.warn(
+          '[react-lds] ExpandableSection:',
+          'Either defaultOpen or open need to be != null.',
+        );
+      }
+    }
+  }
+
+  onClickToggle = (nextOpenState) => {
+    this.setState(() => ({ open: nextOpenState }));
```
fspoettel: If this only toggles with a click, you can simplify this to:

``` js
onClickToggle = () => {
  this.setState(prevState => { open: !prevState.open });
}
```

The added benefit is safety in case you update from outside (react updates are batched)

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,145 @@
...
+
+    return (
+      <span className="slds-truncate slds-p-horizontal_small" title={title}>
+        {title}
+      </span>
+    );
+  }
+
+  renderCollapsable() {
+    const { id, title, open, defaultOpen } = this.props;
```
fspoettel: store open in a const here, you use it multiple times

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,145 @@
...
+        />
+        <span className="slds-truncate" title={title}>
+          {title}
+        </span>
+      </Button>
+    );
+  }
+
+  render() {
+    const { children, className, id, collapsable } = this.props;
```
fspoettel: store open in a const here, you use it multiple times

src/components/ExpandableSection/ExpandableSection.js
```diff
@@ -0,0 +1,145 @@
...
+      </div>
+    );
+  }
+}
+
+ExpandableSection.defaultProps = {
+  className: null,
+  open: null,
+  defaultOpen: null,
+  collapsable: true,
```
fspoettel: might make sense to call this `uncollapsible` and set it to `false` by default, easier to reason about

PR #162, opened 2017-06-27
Feature/update spinner with container

brunottonurb:


src/components/Spinner/Spinner.js
```diff
@@ -5,22 +5,36 @@ import cx from 'classnames';
 import { flavorable } from '../../decorators';
 
 export const Spinner = (props) => {
-  const { className, size, ...rest } = props;
+  const { className, size, container, fixed, ...rest } = props;
```
fspoettel: Wouldn't it be better to have a `<SpinnerContainer />` component that accepts an additional className etc? A use-case for that would be different heights set via a custom class. Also, the `fixed` prop is only relevant if container is set, which is a bit confusing imo. I'm in favor of splitting this into another component

docs/src/app/pages/Spinners/ExampleWithContainerFixed.js
```diff
@@ -0,0 +1,14 @@
...
+import { Spinner, SpinnerContainer } from 'react-lds';
+
+const ExampleWithContainerFixed = () => (
+  <SpinnerContainer>
+    <Spinner size="medium" />
+  </SpinnerContainer>
+);
+
+// the example is not actually fixed, because it would cover the whole page.
+// I could put it in an iframe to contain the effect, but maybe thats over the top
```
philippfrank: Do we need this comment?

src/components/Spinner/Spinner.js
```diff
@@ -14,11 +14,10 @@ export const Spinner = (props) => {
   ];
 
   return (
-    <div className="slds-spinner_container">
-      <div {...rest} className={cx(sldsClasses)} aria-hidden="false" role="alert">
-        <div className="slds-spinner__dot-a" />
-        <div className="slds-spinner__dot-b" />
-      </div>
+    <div {...rest} className={cx(sldsClasses)} role="status">
```
philippfrank: Why not `<div className={cx(sldsClasses)} role="status" {...rest}>`?

That way the user could customize certain props. /cc @fspoettel

fspoettel: @brunottonurb implementation is correct. we destructure the className string from the props and add it to sldsClasses. The role does not need to be overwritten

docs/src/app/pages/Spinners/ExampleWithContainerFixed.js
fspoettel: @brunottonurb it's alright to document the prop and not add an example. remove this example

src/components/Spinner/SpinnerContainer.js
```diff
@@ -0,0 +1,41 @@
...
+  fixed: false,
+};
+
+SpinnerContainer.propTypes = {
+  /**
+   * class name
+   */
+  className: PropTypes.string,
+  /**
+   * needed if dynamically loading portions of component after spinner is showing
```
fspoettel: unclear

PR #163, opened 2017-07-06
Feature/update toast component

brunottonurb:


src/components/Notifications/Notification.js
```diff
@@ -29,9 +30,23 @@ export const Notification = (props) => {
...
 
+  const renderIcon = name =>
+    <Icon
+      className="slds-m-right_small slds-no-flex slds-align-top"
+      sprite="utility"
+      icon={name}
+      size="small"
+    />;
+
+  const renderChildren = content =>
```
fspoettel: `wrapToastContent` is a better name here imo

src/components/Notifications/Notification.js
```diff
@@ -30,23 +30,29 @@ export const Notification = (props) => {
...
+    const iconContainerClasses = [
+      { 'slds-icon_container': true },
+      'slds-m-right_small slds-no-flex slds-align-top',
+    ];
 
-  const renderChildren = content =>
+    return (
+      <span className={cx(iconContainerClasses)}>
+        {icon}
+        <span className="slds-assistive-text" />
```
fspoettel: shouldn't this have a content?

PR #168, opened 2017-07-24
Use title attribute as tooltip for button component

PeterSiegl:


src/components/Button/Button.js
```diff
@@ -29,6 +30,7 @@ export const Button = (props, { cssPrefix }) => {
       onClick={onClick}
       disabled={disabled}
       value={value}
+      title={tooltip}
```
fspoettel: Maybe a fallback to `title` would be a good idea here

PR #175, opened 2017-08-04
Modal: aria-describedby, aria-label, ModalContent: id

brunottonurb:


src/components/Notifications/Prompt.js
```diff
@@ -9,7 +9,7 @@ const Prompt = (props) => {
     <div>
       <Modal {...rest} className={className} label={label} description={description} open={open} dialog prompt>
         <ModalHeader title={title} />
-        <ModalContent>{children}</ModalContent>
+        <ModalContent id={description}>{children}</ModalContent>
```
fspoettel: Can the description contain whitespace? If so, it should not be used here (or encoded)

PR #178, opened 2017-08-23
Add Listbox for pills

fspoettel:


src/components/Pill/Listbox.js
```diff
@@ -0,0 +1,57 @@
...
+    className,
+  ];
+
+  const renderChild = ((child, i) => {
+    const clonedChild = React.cloneElement(child, {
+      'aria-selected': true,
+      tabIndex: i === 0 ? 0 : null,
+    });
+
+    return <li key={i} role="presentation">{clonedChild}</li>;
```
carlo: Every time you put JS and JSX on the same line, a cute kitten drowns in its own tears.

Just saying.

PR #181, opened 2017-08-25
Add File Component

fspoettel:


src/components/File/File.js
```diff
@@ -0,0 +1,190 @@
...
+    hideTitle,
+    image,
+    isLoading,
+    ratio,
+    title,
+    ...rest
+  } = props;
+
+  const renderActions = () => {
+    if (!hideTitle) {
```
philippfrank: I don't get this. Seems it should be `if (hideTitle)`?

src/components/File/File.js
```diff
@@ -0,0 +1,190 @@
...
+    actions,
+    className,
+    externalIcon,
+    fileType,
+    hideTitle,
+    image,
+    isLoading,
+    ratio,
+    title,
+    ...rest
```
philippfrank: Although the linter allows this, can you add a comma to be consistent.

fspoettel: no: https://github.com/eslint/eslint/issues/3627

src/components/File/File.js
fspoettel: the logic is correct, I'll switch the statements tho. That's what I get for renaming `showTitle` to `hideTitle` later in dev

Issue #136, opened 2017-06-09
Update Accessibility and Markup for existing components


fspoettel:
- [x] #191 — Card: Changed `.slds-card__footer` to a `<footer />` element
- [x] #189 — The `PageHeader` component "entity" icon is now contained within `.slds-icon_container`
- [x] #186 — Whenever `.slds-icon-text-default`, `.slds-icon-text-warning`, or `.slds-icon-text-error` is used, it is required to be used in conjunction with `.slds-icon`
- [x] Unless the HTML output is represented as a list — ul, ol, or dl — all examples were updated to use slds-bottom—[position] instead of slds-has-divider—[position]
- [x] Checkbox Group: Removed for attributes on wrapping span elements
- [x] Date Picker: Made the first day cell focusable as it is a ARIA grid widget
- [x] Avatar: Added missing title attributes
- [x] Whenever slds-truncate is used, it is required to be used in conjunction with a [title] attribute
- [x] Date Picker: Added role="dialog" to the slds-datepicker
 - [x] Date Picker: Added aria-label to the dialog to describe its purpose
  - [x] Date Picker: Added aria-multiselectable="true" to the grid to indicate you can selected multiple dates
 - [x] Date Picker: Removed headers attribute from each gridcell
 - [x] Date Picker: Added assistive text for when we mark today's date in the grid. We do this to provide additional meaning since the grey background color is not enough of an indicator to non-sighted users
 - [x] Date Picker: Changed the "Today" option to be a selectable cell, just like any other day in the grid, by removing its <a /> tag
  - [x]  Menus: Added aria-checked to Menu items with role="menuitemcheckbox" in all cases. Shouldn't just be added when true, setting false is important to communicate
 - [x] Modals: Added aria-describedby to the slds-modal element. This points to ID of slds-modal__content
Added aria-label to Header Empty example, in the absence of a header with a heading tag
  - [x] Pills: Add title attributes to all icons or avatars in a Pill
 - [x] Prompt: Added an ID to slds-modal__content for aria-describedby to work on the alertdialog
 - [x] Select: Added missing required attribute to Required and Error states
 - [x] Tabs: Fixed Sub Tabs tabpanels ID typos
- [x] Picklist: Upgrade the ARIA Combobox markup to ARIA 1.1 Specification
Moved role="combobox" attribute from the input element to slds-picklist root element
Moved aria-expanded="true|false" attribute from the input element to slds-picklist root element
Added aria-haspopup attribute to slds-picklist root element
[...]

PR #184, opened 2017-08-28
Multiple updates or fixes datepicker

brunottonurb:
Issue #136 Datepicker-stuff:

> Date Picker: Added role="dialog" to the slds-datepicker
> Date Picker: Added aria-label to the dialog to describe its purpose
> Date Picker: Removed headers attribute from each gridcell
> Date Picker: Added assistive text for when we mark today's date in the grid. We do this to provide additional meaning since the grey background color is not enough of an indicator to non-sighted users
 Date Picker: Changed the "Today" option to be a selectable cell, just like any other day in the grid, by removing its tag
> Date Picker: Made the first day cell focusable as it is a ARIA grid widget

Did not implement the following because AFAIK our Datepicker doesn't support mulitple date input.
> Date Picker: Added aria-multiselectable="true" to the grid to indicate you can selected multiple dates

### DIFF
--- a/src/components/Datepicker/Datepicker.js
+++ b/src/components/Datepicker/Datepicker.js
@@ -361,7 +361,7 @@ export class Datepicker extends React.Component {
     };
 
     return (
-      <td key={dayIndex} className={cx(classes)} headers={day.day()}>
+      <td key={dayIndex} className={cx(classes)}>
         <span className="slds-day" onClick={onClick}>
           {day.date()}
         </span>
--- a/src/components/Datepicker/Datepicker.js
+++ b/src/components/Datepicker/Datepicker.js
@@ -357,14 +357,23 @@ export class Datepicker extends React.Component {
 
     const isToday = day.isSame(moment(), 'day');
 
+    const isSelected = day.isSame(moment(inputValue, defaultDateFormat), 'day');
+
     const classes = {
       'slds-is-today': day.isSame(moment(), 'day'),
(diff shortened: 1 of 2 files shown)


src/components/Datepicker/Datepicker.js
```diff
@@ -354,15 +355,27 @@ export class Datepicker extends React.Component {
...
       'slds-is-today': day.isSame(moment(), 'day'),
       'slds-disabled-text': !inRange,
-      'slds-is-selected': day.isSame(moment(inputValue, defaultDateFormat), 'day'),
+      'slds-is-selected': isSelected,
     };
 
     return (
-      <td key={dayIndex} className={cx(classes)} headers={day.day()}>
+      <td
+        aria-disabled={!inRange ? 'true' : null}
```
carlo: A simple `aria-disabled={!inRange}` should suffice. In the end, we just want to make sure the attribute is set when `inRange` is falsy.

carlo:
@brunottonurb: Sorry, I've meant to approve your changes in the first place. Nonetheless, my comments re `aria-disabled` still stand.

src/components/Datepicker/Datepicker.js
brunottonurb: I just added the null option because in the lds code examples, the aria-disabled attribute does not exist if inRange is truthy. When I set it to false, I get an html attribute, with null, react omits the attribute...

Issue #145, opened 2017-06-09
Update Input with Spinner


fspoettel:
- [x] Added `.slds-has-input-focus` utility class to provide focus styles on an element
- [x] Added an example for Search input with loading spinner - Input with spinner

### PR / Branch

https://github.com/propertybase/react-lds/pull/185

carlo:
PR #185 is open for review.

PR #185, opened 2017-08-29
Update Input with Spinner #145

carlo:
See #145.

### DIFF
--- a/src/components/Form/InputRaw.js
+++ b/src/components/Form/InputRaw.js
@@ -99,19 +99,13 @@ const InputRaw = (props) => {
     return null;
   };
 
-  const renderSpinner = () => {
-    if (!showSpinner) {
-      return null;
-    }
-
-    return (
-      <Spinner
-        brand
-        className="slds-input__spinner slds-m-right_xx-small"
-        size="x-small"
-      />
-    );
-  };
+  const renderSpinner = () => showSpinner && (
+    <Spinner
+      brand
+      className="slds-input__spinner slds-m-right_xx-small"
+      size="x-small"
+    />
+  );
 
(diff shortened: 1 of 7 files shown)


src/components/Form/InputRaw.js
```diff
@@ -91,32 +97,49 @@ const InputRaw = (props) => {
...
+
+    return (
+      <Spinner
+        brand
+        className="slds-input__spinner slds-m-right_xx-small"
+        size="x-small"
+      />
+    );
+  };
+
```
PeterSiegl: even shorter: `return showSpinner && <Spinner ... />`

src/components/Form/InputRaw.js
```diff
@@ -91,32 +97,49 @@ const InputRaw = (props) => {
...
   const sldsClasses = [
-    { 'slds-input_bare': bare },
-    { 'slds-input': !bare },
+    { 'slds-has-input-focus': isFocused },
+    bare ? 'slds-input_bare' : 'slds-input',
```
PeterSiegl: i suppose this should be `slds-input--bare`

carlo: Nope, `.slds-input_bare` is correct, see https://lightningdesignsystem.com/components/input/

carlo:
Thanks for the review, @PeterSiegl.

I'd like @fspoettel to take a look as well — I'd like to make sure I correctly understood his first task in #145 (the focus class).

src/components/Form/InputRaw.js
PeterSiegl: okay, ignore my comment: react-lds is already on 2.3.1

PR #186, opened 2017-08-29
Add new .slds-icon classes where necessary (#136)

carlo:
See #136.


carlo:
Turns out this isn't needed anymore. Closing.

PR #195, opened 2017-08-31
Fix errors in ExpandableSection tests

brunottonurb:


src/components/ExpandableSection/__tests__/ExpandableSection.spec.js
```diff
@@ -4,7 +4,12 @@ import { shallow } from 'enzyme';
 import { Button } from '../../../';
 import { ExpandableSection } from '../ExpandableSection';
 
-const getExpandableSection = (props = {}) => shallow(<ExpandableSection id="heyo" {...props} />);
+const child = <div className="heyo" />;
+const getExpandableSection = (props = {
+  id: 'heyo',
+  children: child,
+  title: 'Title'
+}) => shallow(<ExpandableSection {...props} />);
```
fspoettel: you could also default these by `getComponent = (props = {}) => shallow(<Foo title="A" {...rest}) />`, leaving room to partially apply props via `getComponent({ children: 'bar' })`;

brunottonurb: Ah, I was wondering how to do that!

PR #197, opened 2017-09-01
Updated Accordion to work in controlled fashion

brunottonurb:


docs/src/app/pages/Accordion/ExampleControlled.js
```diff
@@ -0,0 +1,51 @@
...
+  sectionClickHandler = (id) => {
+    this.setState({ openSection: id });
+  }
+
+  render() {
+    const { openSection } = this.state;
+    return (
+      <Accordion
+        open={openSection}
+        sectionClick={this.sectionClickHandler}
```
fspoettel: should be prefixed with `on`

PR #205, opened 2017-09-12
Fix IconSVG background

brunottonurb:
- behavior when no background is specified
- removed superflouous "slds-" in iconsvg background class

### DIFF
--- a/src/components/Icon/IconSVG.js
+++ b/src/components/Icon/IconSVG.js
@@ -6,7 +6,7 @@ const IconSVG = (props, { assetBasePath }) => {
   const { background, className, fill, icon, size, sprite, ...rest } = props;
 
   let backgroundClass = background;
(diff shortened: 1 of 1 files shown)


src/components/Icon/IconSVG.js
```diff
@@ -6,9 +6,10 @@ const IconSVG = (props, { assetBasePath }) => {
   const { background, className, fill, icon, size, sprite, ...rest } = props;
 
   let backgroundClass = background;
-  if (backgroundClass === undefined || backgroundClass === true) {
-    backgroundClass = `slds-${sprite}-${icon}`;
+  if ((backgroundClass === undefined) || (backgroundClass === true)) {
```
fspoettel: let's null this in the defaultProps and check for `=== null`

fspoettel: let's null this in the defaultProps and check for `=== null`
`

PR #206, opened 2017-09-12
Pill without close button

brunottonurb:


src/components/Pill/Pill.js
```diff
@@ -25,17 +25,20 @@ export const Pill = (props) => {
     className
   ];
 
+  const closeButton = onClose === null ? [] :
```
fspoettel: Why an empty array here? Might be `null` as well.

PR #207, opened 2017-09-13
Add read only inputs

brunottonurb:


src/components/Form/Input.js
```diff
@@ -68,6 +81,7 @@ Input.defaultProps = {
   onFocus: () => {},
   onKeyPress: () => {},
   placeholder: null,
+  readOnly: false,
```
fspoettel: null here

src/components/Form/Input.js
```diff
@@ -18,14 +18,34 @@ const Input = (props) => {
...
+  const renderInputRaw =
+    (<InputRaw
+      error={error}
+      errorIcon={errorIcon}
+      iconLeft={iconLeft}
+      iconRight={iconRight}
+      id={id}
+      label={label}
+      placeholder={placeholder}
+      readOnly={readOnly ? true : null}
```
fspoettel: you can check for `readOnly` if you null in the default props, no need for a ternary

src/components/Form/Input.js
```diff
@@ -18,14 +18,34 @@ const Input = (props) => {
...
     required,
     showSpinner,
+    value,
     ...rest,
   } = props;
 
   const hasIconLeft = !!iconLeft || (error && errorIcon);
   const hasIconRight = !!iconRight || !!showSpinner;
 
+  const renderInputRaw =
```
fspoettel: `render` implies a function, this is an element. rename to `inputRaw`

src/components/Form/Input.js
```diff
@@ -18,14 +18,34 @@ const Input = (props) => {
...
+      id={id}
+      label={label}
+      placeholder={placeholder}
+      readOnly={readOnly ? true : null}
+      required={required}
+      showSpinner={showSpinner}
+      value={value}
+      {...rest}
+    />)
+    ;
```
fspoettel: same line please

src/components/Form/Textarea.js
```diff
@@ -55,7 +55,7 @@ const Textarea = (props) => {
 
   return (
     <FormElement required={required} error={error}>
-      <FormElementLabel label={label} id={id} required={required} hideLabel={hideLabel} />
+      <FormElementLabel label={label} id={id} required={required} hideLabel={hideLabel} readOnly={readOnly} />
```
fspoettel: good catch 👍

Issue #167, opened 2017-07-12
Packaging improvements


fspoettel:
- [ ] Provide ES2015 modules to enable tree-shaking in latest webpack
- [ ] Look into providing a support for legacy browsers (IE11, all mobile supported by Sf1)

PR #212, opened 2017-09-28
Add storybook docs, remove decorators

fspoettel:
- [x] Add docs deployment to Github Pages for Storybook. See [storybook-deployer](https://github.com/storybooks/storybook-deployer)
 - [x] Deployment task via Babel & NPM. Provide ES-Modules for Webpack 2.x. Consider upgrading to Babel7 beta #167
 - [x] Remove decorators section in readme
 - [x] Review


fspoettel:
@brunottonurb looking great so far 👍

fspoettel:
@brunottonurb status on deployments?

fspoettel:
@brunottonurb `uglifyjs-plugin` < 1.0 does not support es6 syntax, you need to add a `NODE_ENVIRONMENT=docs` and check for it in babelrc. You can configure the env plugin to add the `uglify`-option, which should compile everything to ES5.

src/components/Button/Button.js
```diff
@@ -97,6 +85,44 @@ Button.propTypes = {
    * adds optional value tag to the button
    */
   value: PropTypes.string,
+  /**
+   * Button flavor: string or strings in array. Flavors: neutral,
```
fspoettel: This is something people will stumble over: Do I use a string or an array? Can I concat multiple flavors in one string? E.g. `icon icon-border`. If I see a string-flavor, I would assume that that is the case. This is a bit counter-intuitive

src/components/Button/Button.js
```diff
@@ -97,6 +85,44 @@ Button.propTypes = {
...
+   brand, destructive, icon, icon-border-filled, icon-container, icon-inverse,
+   icon-border, icon-bare, icon-x-small, reset, success, x-small, small, large
+   */
+  flavor: PropTypes.oneOfType([PropTypes.oneOf([
```
fspoettel: These PropTypes are really verbose, please use a util that outputs them. E.g. `flavorPropType([arrayOfFlavors])`

src/components/Grid/Container.js
```diff
@@ -33,6 +23,29 @@ Container.propTypes = {
    * class name
    */
   className: PropTypes.string,
+  /**
+   * flavor: string or array of strings. Flavors: small, medium, large, x-large,
+   * fluid, left, center, right
+   */
+  flavor: PropTypes.oneOfType([PropTypes.oneOf([
+    'small',
```
fspoettel: I really think we should move sizes to a dedicated prop

src/components/Grid/Grid.js
```diff
@@ -51,8 +33,47 @@ Grid.propTypes = {
...
+    'certical-align-end',
+    'vertical-stretch',
+    'pull-padded',
+    'pull-padded-medium',
+    'pull-padded-large',
+  ]))]),
+  /*
+   * wrap variation
+   */
+  wrap: PropTypes.bool,
```
fspoettel: This makes no sense from a user's perspective. Why is wrap a prop and other flexbox properties not?

src/components/Menu/Menu.js
```diff
@@ -38,11 +37,14 @@ export class Menu extends Component {
       return (
         <Button
           aria-haspopup="true"
-          brand={button.brand}
+          flavor={button.flavor}
+          className={cx([
+            button.brand ? 'slds-button_brand' : null,
```
fspoettel: This would be better solved with an `array.push` and if-statements imo

src/components/Modal/ModalHeader.js
```diff
@@ -30,11 +30,12 @@ const ModalHeader = (props) => {
   };
 
   const isEmpty = !children && !tagline && !title;
+  const theme = prompt === true ? 'error' : prompt;
```
fspoettel: `prompt && 'error'`

src/components/Notifications/Notification.js
```diff
@@ -44,19 +39,19 @@ export const Notification = (props) => {
...
+    </div>
+  );
 
   return (
     <div className="slds-notify_container">
       <div {...rest} className={cx(sldsClasses)} role="alert">
         {icon && wrapIcon(icon)}
         <Button
-          icon-inverse={getThemeName(className) ? undefined : true}
+          flavor={[(theme === 'warning') ? null : 'icon-inverse', 'icon']}
```
fspoettel: See above

src/components/Notifications/Prompt.js
```diff
@@ -49,6 +50,20 @@ Prompt.propTypes = {
    * opens the prompt
    */
   open: PropTypes.bool,
+  /**
+   * slds theme
+   */
+  theme: PropTypes.oneOf([
```
fspoettel: Should use `THEMES` constant

src/components/Popover/Popover.js
```diff
@@ -177,13 +166,42 @@ Popover.propTypes = {
...
-  customLayout: PropTypes.string,
+  customHeaderTheme: PropTypes.oneOf([
+    'warning',
+    'error',
+    'success',
+    'info']),
+  /**
+   * themes: alt-inverse, default, error, info, inverse, offline, shade, success, warning
+   */
+  theme: PropTypes.oneOf([
```
fspoettel: Use THEMES constant

src/components/Popover/Popover.js
```diff
@@ -75,18 +70,19 @@ export class Popover extends React.Component {
   }
 
   renderCloseButton() {
-    const { onClose, className, customLayout } = this.props;
+    const { onClose, customHeaderTheme, theme } = this.props;
+    const invertIcon =
+      customHeaderTheme ? Popover.shouldInvertIcon(customHeaderTheme) : Popover.shouldInvertIcon(theme);
```
fspoettel: `Popover.shouldInvertIcon(customHeaderTheme || theme)`

src/components/Popover/Popover.js
```diff
@@ -23,18 +21,18 @@ export class Popover extends React.Component {
...
+      { 'slds-theme_warning': customHeaderTheme === 'warning' },
+      { 'slds-theme_error': customHeaderTheme === 'error' },
+      { 'slds-theme_success': customHeaderTheme === 'success' },
+      { 'slds-theme_info': customHeaderTheme === 'info' },
     ];
 
     let headerContent;
     let borderRadius;
-    if (!!panels && (typeof customLayout !== 'undefined') && customLayout !== '') {
+    if (!!panels && (typeof customHeaderTheme !== 'undefined') && customHeaderTheme !== '') {
```
fspoettel: I know it's not your code here, but checking for explicit undefined (especially via typeof) is bad practice.

src/components/VerticalNavigation/VerticalNavigation.js
```diff
@@ -1,17 +1,20 @@
...
     prependElement,
-    ...rest,
+    ...rest
   } = props;
 
-  const sldsClasses = ['slds-nav-vertical', className];
+  const flavorClasses =
+    Array.isArray(flavor) ? flavor.map(f => `slds-nav-vertical_${f}`) : `slds-nav-vertical_${flavor}`;
+
+  const sldsClasses = ['slds-nav-vertical', className, flavorClasses];
```
fspoettel: Shouldn't this be `...flavorClasses`?

package.json
```diff
@@ -52,48 +50,42 @@
...
+    "@storybook/react": "^3.2.8",
+    "@storybook/storybook-deployer": "^2.0.0",
+    "babel-cli": "^6.26.0",
+    "babel-core": "^6.26.0",
+    "babel-eslint": "^8.0.1",
+    "babel-jest": "^21.2.0",
+    "babel-plugin-transform-class-properties": "^6.24.1",
+    "babel-plugin-transform-es2015-modules-commonjs": "^6.26.0",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
+    "babel-plugin-transform-react-constant-elements": "^6.23.0",
```
extronics: This introduces some bugs and warnings. For example in `InputRaw`, it will hoist the right `ButtonIcon` out of the `renderIconRight`. As a result, this throws a warning because the icon button  is created in any case, even if there is no icon prop.

Also, it throws a bunch of undefined element warnings: `Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined.`.
Please see https://github.com/babel/babel/issues/5577

Browsing the babel issue tracker for issues with these plugin yields a bunch of results. Might not be a good choice using this optimization.

fspoettel: @extronics 👍 will remove

src/components/Button/Button.js
brunottonurb: Still not sure how to handle this: I could only allow arrays (even for single flavors), or allow concatenated strings as well as single strings and arrays of strings. The problem with concatenated strings is that checking each flavor string for validity with PropTypes is impossible.

src/components/Grid/Container.js
brunottonurb: 👌

src/components/Button/Button.js
fspoettel: Not impossible but would require a custom propType. fwiw, I think arrays are the way to go here (in combination with consolidated props like `size`)

src/components/Grid/Grid.js
brunottonurb: Because wrap is a variation `slds-wrap' and the others are flavors e.g. `slds-grid_align-spread`.

src/components/Modal/ModalHeader.js
brunottonurb: I believe the short circuit wouldn't work the same if prompt is a string like 'info'. I did this to limit breaking changes and still allow prompt theming. Maybe it wasn't a good idea.

fspoettel: Oh, I overlooked the change in PropType for prompt. Disregard the comment

src/components/Notifications/Notification.js
brunottonurb: && returns `falsy`, but I want `null`.

fspoettel: no, you push a needless null key into the array here

PR #213, opened 2017-10-11
Add option for disabling error message display

extronics:
This adds an boolean `errorMessage` flag which can be used, to disable the display of error messages. This allows us to mark (outline) a form element as invalid without showing a message next to it.

### DIFF
--- a/docs/src/app/pages/Form/Checkbox/ExampleErrorMessage.js
+++ b/docs/src/app/pages/Form/Checkbox/ExampleErrorMessage.js
@@ -8,7 +8,7 @@ const ExampleErrorMessage = () =>
       id="checkbox-input-1"
       label="Checkbox Label"
       error="Something is wrong"
-      errorMessage={false}
+      hideErrorMessage
       required
     />
   </div>;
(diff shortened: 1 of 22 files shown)


docs/src/app/pages/Form/Checkbox/ExampleErrorMessage.js
```diff
@@ -0,0 +1,16 @@
...
+
+import { Checkbox } from 'react-lds';
+
+const ExampleErrorMessage = () =>
+  <div>
+    <Checkbox
+      id="checkbox-input-1"
+      label="Checkbox Label"
+      error="Something is wrong"
+      errorMessage={false}
```
PeterSiegl: LGTM! Just a detail: I would call the prop `showErrorMessage` or `hideErrorMessage` to make its purpose clearer.

extronics: Makes sense. I renamed it to hideErrorMessage so it can be used as a boolean flag like `<Input ... hideErrorMessage />`

PR #215, opened 2017-10-13
Additional display props for Datepicker

extronics:
This adds support for adding a custom className to the `Datepicker` component.


PeterSiegl:
Is it possible to hide the error message for the Datepicker?

PR #229, opened 2017-12-11
Fix export of missing component

extronics:
I'm just assuming that the export should be removed, because it isn't exported from `ObjectHome.js` anymore.


brunottonurb:
I think this should be `export class ObjectHomeRaw extends Component { ` in `ObjectHome.js`.
The default export is ObjectHome with clickOutside wrapper and ObjectHomeRaw without wrapper.

extronics:
@brunottonurb done

PR #232, opened 2017-12-12
Fix .babelrc to make exports work in jest

brunottonurb:


extronics:
Background: when linking react-lds to another package, running tests in that other package breaks due to "unknown keyword `export`" errors.
I'm not sure whether this is the right solution - this could be solved in the package running the tests by also transpiling linked dependencies.

extronics:
On the other side, i think it might make sense to not publish this to npm with es6 keywords still in source.

fspoettel:
@extronics 

Right now, we are transpiling all es6 features but the modules because webpack can consume and optimize them. I think it might make sense to export both es6 & commonjs modules in a separate folder like other modules do.

`babel` by default excludes `node_modules` from being processed for performance reasons and the workaround is to adjust that exclude to a negative lookhead like `/node_modules/?!(react-lds)`. I tried that approach for the `pba-react`  but seemingly random things started breaking in other dependencies, I need to investiage this further.

PR #233, opened 2017-12-12
Minor fixes/improvements

brunottonurb:


src/components/Notifications/Notification.js
```diff
@@ -26,19 +26,12 @@ const Notification = (props) => {
...
-
-    return (
-      <span className={cx(iconContainerClasses)}>
-        {icon}
-        <span className="slds-assistive-text">{title}</span>
-      </span>
-    );
-  };
+  const wrapIcon = () => (
+    <span className={'slds-icon_container slds-m-right_small slds-no-flex slds-align-top'}>
```
extronics: This should be plain string prop: `className="slds-icon_container slds-m-right_small slds-no-flex slds-align-top"`

Issue #220, opened 2017-11-06
Lookup fields, unnecessary re-rendering.


extronics:
The lookup fields re-render completely on every click on the page.

PR #246, opened 2018-01-04
Lookup: fix unnecessary rerender on clickoutside

brunottonurb:
Issue #220

I think the rerenders are triggered by enhanceWithClickOutside and the lack of a shouldComponentUpdate function.

I put in the shallow comparison func for props and state from the [legacy pureRenderMixin](https://developmentarc.gitbooks.io/react-indepth/content/life_cycle/update/using_should_component_update.html)

If this is an acceptable solution I will put `shallowCompare` into `utils`.

### DIFF
--- a/src/components/Lookup/Lookup.js
+++ b/src/components/Lookup/Lookup.js
@@ -61,7 +61,10 @@ const propTypes = {
   /**
    * label for the input field in the lookup component
    */
-  inputLabel: PropTypes.string.isRequired,
+  inputLabel: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.node,
+  ]).isRequired,
   /**
    * label for the dropdown in the lookup component
    */
--- a/src/components/Lookup/Lookup.js
+++ b/src/components/Lookup/Lookup.js
@@ -5,6 +5,7 @@ import enhanceWithClickOutside from 'react-click-outside';
 import debounce from 'lodash.debounce';
 import omit from 'lodash.omit';
 
+// import shallowCompare from '../../utils';
 import {
   FormElement,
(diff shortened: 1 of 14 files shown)


brunottonurb:
@extronics I wasn't sure if I could do that because of

> Furthermore, React.PureComponent’s shouldComponentUpdate() skips prop updates for the whole component subtree. Make sure all the children components are also “pure”.

extronics:
@brunottonurb should be no issue in this case because all children of Lookup are functional components (which are pure).

brunottonurb:
After doing some fiddling around, I think using PureComponent is the way to go: The case where a rerender is triggered needlessly by clickOutside is when the prop `open` did not change. A custom `shouldComponentUpdate` function would essentially do the same thing as the PureComponent shallowCompare.

src/components/Lookup/Lookup.js
```diff
@@ -1,10 +1,11 @@
-import React, { Component } from 'react';
+import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import enhanceWithClickOutside from 'react-click-outside';
 import debounce from 'lodash.debounce';
 import omit from 'lodash.omit';
 
+// import shallowCompare from '../../utils';
```
fspoettel: Remove

src/components/Lookup/Lookup.js
```diff
@@ -218,6 +219,45 @@ export class LookupRaw extends Component {
     }
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
```
fspoettel: Remove

PR #248, opened 2018-01-18
Feature/button with a tag and href

brunottonurb:


src/components/Button/Button.js
```diff
@@ -5,9 +5,11 @@ import { applyDecorators, decoratorProp } from '../../utils';
 
 const Button = (props) => {
   const {
+    a,
```
extronics: IMHO this is redudant. You could just assume an `a` tag if the `href` prop is present.

PR #253, opened 2018-01-24
Update Badge for 2.5.0

fspoettel:


src/components/Badge/Badge.js
```diff
@@ -2,35 +2,47 @@ import React from 'react';
...
-  /** SLDS Theme */
+  /**
+   * Badge flavor: array of flavors, you can also provide a single flavor string.
+   * Flavors: inverse, lightest
+  */
+  flavor: decoratorProp([
+    'inverse',
+    'lightest'
+  ]),
+  /** DEPRECTATED: SLDS Theme */
```
extronics: With the breaking changes already introduced, what about removing the theme immediately instead of deprecating?

fspoettel: We still need it for colored badges that we are currently using, not possible with new badges

src/components/Badge/IconOnlyBadge.js
```diff
@@ -0,0 +1,19 @@
...
+);
+
+IconOnlyBadge.propTypes = {
+  /** The Icon to display. Can also be a string for notification-badges */
+  children: PropTypes.oneOfType([PropTypes.element, PropTypes.string]).isRequired,
+  /** Description of the Icon */
+  title: PropTypes.string.isRequired,
+};
+
+export default IconOnlyBadge;
```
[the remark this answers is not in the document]
fspoettel: Good point. After creating this cmp, I noticed that Salesforce is inconsistent here and has a few badges in other component that need assistive-text and do not contain icons (despite the docs saying otherwise). Will change

PR #259, opened 2018-01-25
Add Carousel

extronics:


stories/Carousel.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
```
fspoettel: please limit the width of the module, the huge image looks a bit broken

src/components/Carousel/Carousel.js
```diff
@@ -0,0 +1,258 @@
...
+      this.updatePanels(nextChildren);
+    }
+
+    if (!!nextAutoPlayActive !== prevAutoPlayActive) {
+      if (nextAutoPlayActive) this.startAutoPlay();
+      else this.stopAutoPlay();
+    }
+  }
+
+  onClickIndicator = e => this.setActivePanel(Number(e.target.dataset.index));
```
fspoettel: This should also stop or reset the autoplay when active. currently, there is an edge-case where two indicators appear to be active when autoplay is active and an item is clicked, see attached screenshot.

screen shot 2018-01-25 at 17 05 12

stories/Carousel.js
extronics: The carousel is designed to fill it's container. I limited the storybook example the same way the demo on lightningdesignsystem.com is limited.

src/components/Carousel/Carousel.js
extronics: This was actually an issue with the indicator keeping its `:focus` state. I now added a blur() to the click event, not sure though whether this is the best solution ...

stories/Carousel.js
fspoettel: Salesforce applies `width: 480px` in a wrapper

fspoettel: Edit: nvm misread your comment. All good 👍

PR #272, opened 2018-02-06
Feature/grid order prop

brunottonurb:
Add non-responsive `order` prop to reorder `Columns` in `Grid`. Responsive usage is similar to `sizeOf ` prop, e.g. `small-order="3"` or `medium-order="4"`.

### DIFF
--- a/src/components/Grid/Column.js
+++ b/src/components/Grid/Column.js
@@ -13,8 +13,9 @@ const validBreakpoints = [
   'max-large',
 ];
 
-const breakPointProp = breakpoint => `${breakpoint}-sizeOf`;
+const breakPointProp = (breakpoint, suffix) => `${breakpoint}-${suffix}`;
 const sizeRegex = /^([1-9]|1[0-2])-([1-9]|1[0-2])$/;
+const orderRegex = /^((1[0,1,2])|[1-9])$/;
 
 const Column = (props) => {
   const {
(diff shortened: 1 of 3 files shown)


stories/Grid.js
```diff
@@ -102,6 +102,46 @@ stories
...
+        sizeOf="1-2"
+        order="1"
+        small-order="2"
+        medium-order="3"
+      >
+        <Box>One. Try adjusting the window width!</Box>
+      </Column>
+      <Column
+        className="slds-m-bottom--large slds-p-horizontal_small"
+        sizeOf="1-2"
```
fspoettel: You could go with a `medium-sizeOf` here, the demo looks broken when resizing

src/components/Grid/Column.js
```diff
@@ -13,8 +13,9 @@ const validBreakpoints = [
   'max-large',
 ];
 
-const breakPointProp = breakpoint => `${breakpoint}-sizeOf`;
+const breakPointProp = (breakpoint, suffix) => `${breakpoint}-${suffix}`;
 const sizeRegex = /^([1-9]|1[0-2])-([1-9]|1[0-2])$/;
+const orderRegex = /^((1[0,1,2])|[1-9])$/;
```
fspoettel: the regex is rather simple in what it does, maybe a functional condition with Math.min/Math.max would be better here? You could keep the regex for consistency's sake, but the function will be more performant.

src/components/Grid/Column.js
```diff
@@ -36,18 +37,27 @@ const Column = (props) => {
...
+    const orderString = breakpoint === '' ? 'order' : `${breakpoint}-order`;
+    const order = props[orderPropName];
+    if (orderRegex.test(order)) {
+      const position = parseInt(order.split('-')[0], 10);
+      sldsClasses.push(`slds-${orderString}_${position}`);
+    }
   });
 
-  const restProps = omit(rest, validBreakpoints.map(breakPointProp), 'sizeOf');
+  let restProps = rest;
```
fspoettel: You could remove one Iteration and the triple assignment by changing to:

``` js
const generatedProps = [];

validBreakpoints.forEach((a) => {
  generatedProps.push(
    breakPointProp(a, 'sizeOf')), 'sizeOf'),
    breakPointProp(a, 'order')), 'order'),
  );
});

const restProps = omit(rest, generatedProps)
```

src/components/Grid/Column.js
```diff
@@ -36,18 +37,27 @@ const Column = (props) => {
...
-
-    const size = props[propName];
+    const size = props[sizePropName];
     if (sizeRegex.test(size)) {
       const from = parseInt(size.split('-')[0], 10);
       const to = parseInt(size.split('-')[1], 10);
       sldsClasses.push(`slds-${sizeString}_${from}-of-${to}`);
     }
+
+    const orderPropName = breakpoint === '' ? 'order' : breakPointProp(breakpoint, 'order');
```
fspoettel: `breakpoint === ''` can be stored in a const here.

src/components/Grid/Column.js
brunottonurb: The prop can be a string as well as a number. I could use parseInt and isNaN to check the prop. Are regex really that slow?

fspoettel: We can keep it here, but in general, JS regexes are slower by several orders of magnitude than other operations

PR #273, opened 2018-02-06
New Tabs implementation

fspoettel:
## Changes

- `Tabs` has a new signature instead of an array of `tabs`, it now uses an array of `children`. This makes it easier to integrate & optimize (no arrays in render)
- Keyboard support as outlined in the LDS docs
- It is possible to not render the subtrees of empty tabs. This is useful in order to not trigger expensive init logic in hidden subtrees

## Things to watch out for when reviewing

- Keyboard logic needs refs
- I managed to get each tab-change down to one re-render by using a PureComponent and optimizing the event phases in the focus/blur handlers. Please check if the comments added are sufficient

## Todo

- [x] Add controlled tabs
- [x] Use id to identify tabs instead of indexes
- [x] ~~Maybe add subtabs~~ No time right now, should be easy enough though
- [x] Add specs


fspoettel:
@brunottonurb I tried out a new way to handle Uncontrolled/Controlled components. The `Tabs` components is implemented as a light, stateful wrapper around the `ControlledTabs` which are exported as well. What do you think about this approach?

src/components/Tabs/ControlledTabs.js
```diff
@@ -0,0 +1,211 @@
...
+      const isActive = id === activeTab;
+      const hideTab = !isActive && !renderInactiveTabs;
+
+      return React.cloneElement(child, {
+        'aria-labelledby': getAriaLabel(child.props.id),
+        isActive: hideTab ? null : isActive,
+        scoped
+      });
+    };
+  }
```
extronics: These functions like `getActiveTabRenderer` or `getOnBlur` don't curry/memoize anything. Why wrap them into a getter? For `this` binding?

src/components/Tabs/ControlledTabs.js
```diff
@@ -0,0 +1,211 @@
...
+
+  getOnLinkFocus(id) {
+    return () => { this.setState({ focusedTab: id }); };
+  }
+
+  getOnBlur() {
+    return ({ currentTarget }) => {
+      /** Workaround to only fire blur when the whole navigation is unfocused */
+      setTimeout(() => {
```
extronics: This workaround doesn't seem to make a big difference to me (there is an extra setState() but it doesn't seem to cause a re-render, probably merged with the `setState()` in `getOnLinkFocus` afterwards or so ...).

extronics:
The focus cycling for the tab links is pretty far off if the controlled tabs are used without a proper `onChangeTab` implementation. Not sure whether it's worth improving this, as it affects mainly the storybook demo ...

src/components/Tabs/ControlledTabs.js
fspoettel: Firefox? It's working in Safari & Chrome for me, Firefox has a 10 yo bug relating to this :/

fspoettel:
I noticed this too in the demo. Not sure if improving the demo is worth it because if wrapped, the usage details are hidden by the code demo block. I think I'll add an info section to clear this up

src/components/Tabs/ControlledTabs.js
extronics: Yes, i was testing in Firefox. Removing the setTimeout() didn't make a difference there so i was wondering whether i was missing some edge-case ...

src/components/Tabs/ControlledTabs.js
fspoettel: They used to curry an index, I'll rewrite them.

src/components/Tabs/Tab.js
```diff
@@ -0,0 +1,73 @@
...
+    </div>
+  );
+};
+
+Tab.defaultProps = {
+  className: null,
+  isActive: false,
+  renderTitle: null,
+  scoped: false,
+  tabTitle: null,
```
brunottonurb: This is not used anywhere, is it? Probably a leftover.

fspoettel: right, will fix

src/components/Tabs/Tab.js
```diff
@@ -0,0 +1,72 @@
...
+      {...rest}
+      aria-labelledby={getAriaLabel(id)}
+      className={cx(
+        getTabsClass('__content', scoped),
+        isActive ? 'slds-show' : 'slds-hide'
+      )}
+      id={id}
+      role="tabpanel"
+    >
+      {isActive === null ? null : children}
```
extronics: This is a somewhat confusing double-use of `isActive`. I'd make this two props: `isActive` for controlling the style, `renderWhenInactive` to control the hidden rendering behaviour.

fspoettel: I tried to minimize the amount of internally used props in the user-facing API Tab `propTypes`. I figured that this is ok to do since I have full control over these props here and added specs for it.

PR #274, opened 2018-02-07
Fluid carousel layout

extronics:
- This moves the keyboard event handler to the anchor which is supposed to receive the keyboard events
- Also, this adds a way to render the panel image as a background image, thus supporting custom styles with `background-size` and `background-position`

### DIFF
--- a/src/components/Carousel/CarouselPanel.js
+++ b/src/components/Carousel/CarouselPanel.js
@@ -5,20 +5,39 @@ import cx from 'classnames';
 const CarouselPanel = (props) => {
   const {
     active,
+    backgroundStyle,
     children,
     className,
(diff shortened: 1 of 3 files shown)


src/components/Carousel/CarouselPanel.js
```diff
@@ -73,6 +102,11 @@ CarouselPanel.propTypes = {
    */
   imageUrl: PropTypes.string.isRequired,
 
+  /**
+   * Callback for keyboard interaction on an active panel.
+   */
+  onKeyboardInteraction: PropTypes.func,
```
fspoettel: You could require this here and mark it as private as it should not be messed with by end users

PR #283, opened 2018-02-10
Add FocusTrap & Keyboard closing to Modal

fspoettel:


src/components/Modal/Modal.js
```diff
@@ -1,111 +1,113 @@
...
+    return (
+      <div>
+        <section
+          {...rest}
+          className={cx(
+            'slds-modal',
+            { [`slds-${transitionStyle}`]: open },
+            { 'slds-modal_prompt': !!prompt },
+            className
+          )}
```
philippfrank: Please extract the classes into a const.

src/components/Modal/Modal.js
```diff
@@ -1,111 +1,113 @@
...
+  transitionStyle: PropTypes.oneOf([
+    'fade-in-open',
+    'slide-up-saving',
+    'slide-up-open',
+    'slide-down-cancel'
+  ]),
   children: PropTypes.node.isRequired,
-  /**
-   * class name
-   */
```
philippfrank: Why did you remove the comments?

src/components/Modal/Modal.js
```diff
@@ -1,111 +1,113 @@
...
+    const name = child.type.displayName || child.type.name;
 
-    if (childName === 'ModalHeader') {
-      return React.cloneElement(child, {
-        key: label,
-        label,
-        prompt,
-        uncloseable: child.props.uncloseable !== undefined ? child.props.uncloseable : !!prompt,
-      });
+    if (name === 'ModalContent') {
```
philippfrank: Is something like `if (child.type === ModalHeader) { ... }` possible (without the hardcoded component name)?

src/components/Modal/Modal.js
fspoettel: still need to add those again

src/components/Modal/Modal.js
fspoettel: Sorta, but checking for a string is more performant here

philippfrank: Why?

fspoettel: Because a React Component is an object and shallow comparing objects of unknown shape is more expensive than comparing two strings.

philippfrank: Alright, I thought that `ModalHeader` would be a function and that `child.type` would be exactly that function, but you're probably right ✅

PR #284, opened 2018-02-12
Add size property to Picklist component

quadreex:


philippfrank:
@quadreex I know we have this JIRA ticket number convention in the style guide but I'd say since this is an open source repo and the outside world cannot associate these numbers - we skip them here.

src/components/Form/Picklist.js
```diff
@@ -120,6 +123,8 @@ export class PicklistRaw extends Component {
   }
 }
 
+PicklistRaw.size = PICK_LIST_SIZE;
```
fspoettel: Why do we need to set this here? There is a lot of indirection going on with the `constants` file and the static property here

src/components/Form/constants.js
```diff
@@ -0,0 +1,13 @@
+import PropTypes from 'prop-types';
+
+export const PICK_LIST_SIZE = {
+  SMALL: 'small',
```
fspoettel: These are pre-defined slds terms, no need to hide them behind constants.

quadreex: `PICK_LIST_SIZE` itself is a constant to hide slds internals.

We will use it as:
```
<Picklist
...
size={Picklist.size.SMALL}
```

src/components/Form/Picklist.js
quadreex: We need this static property to hide slds internals from the user of our component:
```
<Picklist
...
size={Picklist.size.SMALL}
>
```

fspoettel: We do not do this in any other place and this is not necessary as Salesforce does not change these terms and they are also used all over the place in `slds`-classes, which our users (us) have to use all the time. This introduces a new concept that is not documented well and only applied in this component.

fspoettel: If you think that this should be changed, please open a ticket and we can discuss this as a team. Going forward and changing this in one place while leaving all other components as is is not ideal

PR #285, opened 2018-02-12
Prevent Tabs componentWillReceiveProps from always triggering

fspoettel:
@extronics I think you might run into the same with Carousel. Comparing react children always results in true


extronics:
@fspoettel thanks for the hint, will have a look into this.

Issue #147, opened 2017-06-09
Add Combobox alias for Lookups


fspoettel:
- [ ] Lookup: Uses new combobox HTML
- [x] Added Combobox component — Combobox

fspoettel:
Still a prototype

PR #289, opened 2018-02-13
Add Combobox

extronics:
Adds support for the variants (except multi entity) as described in https://lightningdesignsystem.com/components/combobox/

Can be (and should become) the base for all the other combobox-based components (`Picklist`, `Lookup`) etc.


fspoettel:
#147 #142 #188

fspoettel:
@extronics what is the status here?

extronics:
@fspoettel i've put it on hold while i'm working on more urgent topics. Is there a current need for this component?

fspoettel:
Nope, just out of curiosity

stories/Form.js
```diff
@@ -146,6 +148,61 @@ stories
       size={select('Size', ['', 'small', 'medium', 'large'], '') || undefined}
     />
   ))
+  .add('Combobox', () => (
+    <Combobox
+      inlineListbox={boolean('Inline listbox variant', false)}
+      readOnly={boolean('Read-only variant', false)}
+      error={text('Error', '') || undefined}
+      height={select('Height', [5, 7, 10]) || undefined}
```
PeterSiegl: Returns a `string` but should return a `number` for the height.

src/components/Form/Combobox.js
```diff
@@ -0,0 +1,333 @@
...
+  getSelectHandler = itemKey => memoize(() => this.onSelect(itemKey));
+
+  getSelectedItems() {
+    const { items } = this.props;
+
+    return items.filter(item => item.selected === true);
+  }
+
+  handleClickOutside() {
+    this.setState({ open: !this.state.open });
```
PeterSiegl: Can be removed in favor of `onClickOutside`.

src/components/Form/Combobox.js
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  renderInput() {
+    const {
+      id,
```
PeterSiegl: pls adjust order of vars.

src/components/Form/Combobox.js
```diff
@@ -0,0 +1,333 @@
...
+
+export class ComboboxRaw extends Component {
+  static propTypes = propTypes
+
+  static defaultProps = {
+    className: null,
+    closeOnClickOutside: false,
+    closeOnSelect: true,
+    error: null,
+    height: null,
```
PeterSiegl: I would set a default height e.g. `5` or `7` as we always run into problems with long dropdowns.

src/components/Form/Combobox.js
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  renderInlineListbox() {
+    const selectedItems = this.getSelectedItems();
+
+    return (
+      <Listbox className="slds-listbox_inline">
+        {selectedItems.map(({ icon, key, label }) => (
+          <Pill
+            icon={icon ? <Icon icon={icon.icon} size="small" sprite={icon.sprite} /> : null}
```
PeterSiegl: Could be moved to a function.

src/components/Form/Combobox.js
```diff
@@ -0,0 +1,333 @@
...
+  renderSelectedPills() {
+    const { onSelect } = this.props;
+
+    const selectedItems = this.getSelectedItems();
+
+    return (
+      <Listbox className="slds-listbox_inline slds-p-top_xxx-small">
+        {selectedItems.map(({ icon, key, label }) => (
+          <Pill
+            icon={icon ? <Icon icon={icon.icon} size="small" sprite={icon.sprite} /> : null}
```
PeterSiegl: Could be moved to a function, see above.

src/components/Form/ComboboxDropdownListItem.js
```diff
@@ -0,0 +1,134 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
+import { IconSVG, MediaObject } from '../../';
+
+const ComboboxDropdownListItem = (props) => {
+  const {
+    children,
```
PeterSiegl: pls adjust order of vars.

src/components/Form/ComboboxDropdownListItem.js
```diff
@@ -0,0 +1,134 @@
...
+   */
+  onClick: PropTypes.func,
+  /**
+   * sets this item into a selection state that displays the icon
+   */
+  selected: PropTypes.bool,
+};
+
+ComboboxDropdownListItem.defaultProps = {
+  className: null,
```
PeterSiegl: pls adjust order of vars.

src/components/Form/ControlledCombobox.js
```diff
@@ -0,0 +1,95 @@
...
+
+  render() {
+    const { ...rest } = this.props;
+
+    const { items, value } = this.state;
+
+    return (
+      <Combobox
+        {...rest}
+        onAdd={this.onAdd}
```
PeterSiegl: pls adjust order of vars.

src/components/Form/Combobox.js
fspoettel: @PeterSiegl I disagree, this should not be a function

src/components/Form/Combobox.js
fspoettel: I disagree, this should not be a function

src/components/Form/Combobox.js
```diff
@@ -0,0 +1,333 @@
...
+        role="textbox"
+        value={label}
+      />
+    );
+  }
+
+  renderComboboxItems() {
+    const { items } = this.props;
+
+    return items.map(item => (
```
fspoettel: `item` could be destructured

extronics:
@PeterSiegl "Inline listbox has doubled border." <= this is an issue with the CSS, IMHO we are doing the right thing here:
SLDS adds the border around the input if it is inside a inline listbox and has a value. This works for their code samples, because they don't have the `value` attribute on their `input`. But it doesn't work for react (or any code which sets the `value` attribute).
So i'd suggest we leave it like that and leave the visual fix to users of the package who include their own, custom, css.

src/components/Form/Combobox.js
```diff
@@ -0,0 +1,333 @@
...
+    ));
+  }
+
+  renderInlineListbox() {
+    const selectedItems = this.getSelectedItems();
+
+    return (
+      <Listbox className="slds-listbox_inline">
+        {selectedItems.map(({ icon, key, label }) => (
+          <Pill
```
fspoettel: `renderInlineListBox` and `renderSelectedPills` are pretty much identical and could be DRYed up with little effort

src/components/Form/Combobox.js
```diff
@@ -0,0 +1,333 @@
...
+      size,
+    } = this.props;
+
+    const { open } = this.state;
+
+    const condition = closeOnClickOutside && open;
+
+    const renderPills = !inlineListbox && (
+      (readOnly && this.getSelectedItems().length > 1) ||
+      (!readOnly && this.getSelectedItems().length > 0)
```
fspoettel: This is the third time `getSelectedItems` is computed, it might make sense to store in a const

stories/Form.js
```diff
@@ -146,6 +148,61 @@ stories
...
+      id="picklist-1"
+      isDisabled={boolean('Disabled', false) || undefined}
+      hideLabel={boolean('Label hidden', false) || undefined}
+      required={boolean('Required', false) || undefined}
+      items={object('Items', [
```
fspoettel: I do not understand how the uncontrolled `Combobox` should work here. This is supposed to be a _seed_, correct? Shouldn't the Combobox handle all selections from then on? Currently, it does not seem like this is the case.

stories/Form.js
extronics: This is a limitation of the storybook knobs addon. Can't do much about it at the moment.

stories/Form.js
extronics: The uncontrolled `Combobox` is supposed to only render what's provided via prop, without managing some internal state. So `items`is not really a seed but just what needs to be rendered.

src/components/Form/ControlledCombobox.js
```diff
@@ -0,0 +1,95 @@
+import React, { Component } from 'react';
```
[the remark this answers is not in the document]
extronics: I think this is just a question how we define what "controlled" means. Does it mean that I, as a user of this library, am controlling the state of the component via prop? Or is the component controlling a wrapped component? My thinking was the later but i'd be ok with the first too. In this case, we'd have to swap the names.
As for the `...Raw`: this only affects the `closeOnClickOutside` property (which only exists for backwards compatibility with the `Picklist`) and can always be overwritten when using the `Combobox`. So instead of creating a `ControlledComboboxRaw` one could always just do a `<ControlledCombobox closeOnClickOutside={false} ...>`.

stories/Form.js
fspoettel: The naming is backwards then. You are describing `ControlledCombobox` since it it controlled by props.

src/components/Form/ControlledCombobox.js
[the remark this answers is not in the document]
fspoettel: See https://reactjs.org/docs/forms.html#controlled-components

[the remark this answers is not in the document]
fspoettel: Yes that would work. We already have some `Controlled` components though, it would probably better to have this consistent and either rename all of them or stick with `Controlled`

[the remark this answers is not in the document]
extronics: I worked on the naming and also got rid of the `...Raw` component, because it's only purpose was to provide a default for the `ClickOutside` component.

src/components/Combobox/__tests__/ControlledCombobox.spec.js
```diff
@@ -68,11 +68,18 @@ describe('</Combobox />', () => {
...
-  it('is closed by default and opens when the input was clicked', () => {
+  it('is closed by default and opens when the input was clicked', (done) => {
     const picklist = mounted.find('div.slds-combobox').first();
     expect(picklist.hasClass('slds-is-open')).toBeFalsy();
     picklist.find('input').simulate('click');
-    expect(picklist.hasClass('slds-is-open')).toBeTruthy();
+
+    setTimeout(() => {
+      try {
+        expect(picklist.hasClass('slds-is-open')).toBeTruthy();
```
fspoettel: This test is never executed

PR #290, opened 2018-02-13
Add Slider

quadreex:


extronics:
Small layout glitch: the "required" asterisk should be on the same line as the label:

image

src/components/Form/__tests__/Slider.spec.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import { getUniqueHash } from '../../../utils';
```
extronics: Please remove, yields an linter error

quadreex:
@extronics @fspoettel Does it really make sense to have `required` flag for this component, as we always have value for it?
Maybe I'd better remove this flag?

extronics:
@quadreex you're right. Also, the salesforce design system doesn't seem to provide a `required` state for the slider at all.

PR #291, opened 2018-02-14
New buttons implementation

fspoettel:
- [x] Split buttons into clean components
- [x] Add stories for new components
- [x] Add back stories for menu buttons & buttongroups
- [x] Add specs
- [x] Update dependent components


philippfrank:
@fspoettel These are very good enhancements. I'd like to throw something in here:

This is the current syntax for a simple button with an icon. 

```jsx
<Button flavor="neutral" onClick={action()}>
  <ButtonIcon
    position="left"
    sprite="utility"
    icon="download"
  />
  Download
</Button>
```

I would like to write less code (with the option to still write exactly the above if needed). What about if we'd (optionally) offer something like:

```jsx
// default flavor
// default icon position
<Button onClick={action()} icon="utility.download">
  Download
</Button>
```

fspoettel:
@philippfrank very good idea! Now that we have a dedicated `IconButton` and I solved the one case where this was not possible in the past (`more` prop for IconGroups), it is safe to add this. I'm a bit on the fence whether we should go with `icon="" sprite=""` or `icon="sprite.icon"` as you proposed since it would be the only place in the whole project where that is possible. We could implement `<IconButton icon="x" sprite="y" />` first as a shortcut and later implement your proposed syntax across all components. What do you think?

philippfrank:
@fspoettel I am very open to anything which requires less code ;-)

fspoettel:
@philippfrank the minimal signatures now are:

``` jsx
/* Same content/title */
<Button title="Bleep" />

/* Differing content/title */
<Button title="Bleep">Bloop</Button>

/*  Same content/title and left icon */
<Button title="Bleep" icon="foo" sprite="bar" />

/*  Differing content/title and left icon */
<Button title="Bleep" icon="foo" sprite="bar">Bloop</Button>

/* Right Icon */
<Button title="Bleep">
  <ButtonIcon
    position="right"
    icon="foo"
    sprite="bar"
  />Bleep
</Button>
/* Icon-Only button */
<IconButton icon="foo" sprite="bar" />

/* Stateful Icon-Only button */
<StatefulIconButton selected icon="foo" sprite="bar" />
```

The obvious worst-case is a button with a right-aligned icon. We could add a shortcut prop `position` and default it to `left` not sure whether that is worth it though. cc @brunottonurb

src/components/Button/Base/Button.js
```diff
@@ -0,0 +1,105 @@
...
+      )}
+      {!isRightShortcut ? content : null}
+    </ButtonEl>
+  );
+};
+
+Button.defaultProps = {
+  children: null,
+  className: null,
+  flavor: 'neutral',
```
philippfrank: 👍

src/components/Button/Base/Button.js
```diff
@@ -0,0 +1,105 @@
...
+  href: null,
+  icon: null,
+  iconPosition: null,
+  sprite: null,
+  title: null,
+};
+
+Button.propTypes = {
+  /**
+   * Used to set content. Can be a string, a `ButtonIcon` or a combination of those
```
philippfrank: Technically it `can` be everything. I'd remove that addendum -- it implies specific child types.

src/components/Button/Base/Button.js
```diff
@@ -0,0 +1,105 @@
...
+   */
+  flavor: PropTypes.oneOf([
+    'neutral',
+    'brand',
+    'destructive',
+    'success',
+    'inverse'
+  ]),
+  /**
+   * Optional href, renders as `a` when set.
```
philippfrank: Remove interpunction

src/components/Button/Base/StatefulButtonState.js
```diff
@@ -0,0 +1,66 @@
...
+    state,
+    title,
+    ...rest
+  } = props;
+
+  /* eslint-disable no-nested-ternary */
+  const selectedClass = state === 'selected'
+    ? 'slds-text-selected'
+    : state === 'focus' ? 'slds-text-selected-focus' : 'slds-text-not-selected';
+  /* eslint-enable */
```
philippfrank:
```es6
let selectedClass = null;
switch (state) {
  case 'selected':
    // ...
  case 'focus':
    // ...
  default:
    // ... 
}
```

philippfrank:
Awesome enhancements @fspoettel 💯

PR #292, opened 2018-02-15
Carousel: only update panels if children actually changed

extronics:
Same as in #285 

I implemented the check based on the child props (and length) to ensure the state is updated when child keys, number of childs or other things (image, title, description) change.


src/components/Carousel/Carousel.js
```diff
@@ -33,7 +34,10 @@ class Carousel extends Component {
     const { children: prevChildren } = this.props;
     const { autoPlayActive: prevAutoPlayActive } = this.state;
 
-    if (nextChildren !== prevChildren) {
+    const childrenChanged = nextChildren.length !== prevChildren.length ||
```
fspoettel: It is a good idea to check for length change first, I will amend that in #285 .

src/components/Carousel/Carousel.js
```diff
@@ -33,7 +34,10 @@ class Carousel extends Component {
...
+      !isEqual(nextChildren.map(({ props }) => props), prevChildren.map(({ props }) => props));
```
fspoettel: What do you think about working with `Array.every()` and `Object.values()` here? It could potentially improve the legibility of this line

Issue #144, opened 2017-06-09
Rework Menus


fspoettel:

fspoettel:
@brunottonurb please talk to @peter or me before starting to work on this. We need to make sure we get this right

brunottonurb:
🆗

PR #293, opened 2018-02-28
Rework menus #144

brunottonurb:
- ~~Not sure about how to implement the single event listener per menu. It works, but I don't think it's good.~~
- ~~I can't get the unit test to work on it: simulate('click') is not a function~~

> Even though the name would imply this simulates an actual event, .simulate() will in fact target the component's prop based on the event you give it. For example, .simulate('click') will actually get the onClick prop and call it.

- ~~ObjectHome still needs fixing with the new Menu~~


src/components/Menu/ControlledMenu.js
```diff
@@ -0,0 +1,209 @@
...
+   * ```
+   * {
+   *    icon: 'settings',
+   *    sprite: 'utility',
+   *    title: 'Click me',
+   *    noBorder: true,
+   * }
+   * ```
+   */
+  button: PropTypes.shape({
```
fspoettel: At this point and with #291 landing, it would make more sense to always pass a button imo.

src/components/Menu/MenuSubHeader.js
```diff
@@ -0,0 +1,43 @@
...
+const MenuSubHeader = (props) => {
+  const { children, className, divider, ...rest } = props;
+  const sldsClasses = [
+    'slds-dropdown__header',
+    'slds-truncate',
+    { 'slds-has-divider_top-space': divider },
+    className,
+  ];
+  return (
+    <li className={cx(sldsClasses)} title={children} role="separator" {...rest}>
```
fspoettel: Split into multiple lines

src/components/Menu/MenuItem.js
```diff
@@ -62,8 +61,8 @@ const MenuDropdownListItem = (props) => {
 
   return (
     <li {...rest} className={cx(sldsClasses)} role="presentation">
-      <a role={role} onClick={onClick} aria-checked={selected}>
-        <span className="slds-truncate" title={children}>
+      <a role={role} aria-checked={selected} id={id}>
+        <span className="slds-truncate" title={children} id={id}>
```
fspoettel: Duplicate ids are invalid in html, one of these should be removed

stories/Menu.js
```diff
@@ -40,48 +40,60 @@ stories
...
       disabled={boolean('Disabled', false)}
-      isOpen={boolean('isOpen', false)}
       last={boolean('Last', false)}
-      nubbin={boolean('Nubbin', false)}
+      nubbin={boolean('Nubbin', true)}
       position={select('Position', [
         '', 'top-left', 'top', 'top-right', 'bottom-left', 'bottom', 'bottom-right'
       ], '') || undefined}
       size={select('Size', ['', 'small', 'medium', 'large'], '') || undefined}
+      isOpen={boolean('isOpen', false) || undefined}
```
fspoettel: Should be renamed to `defaultOpen` for the uncontrolled menu since it should not do anything after the first click

src/components/Menu/ControlledMenu.js
brunottonurb: Okay

src/components/Menu/ControlledMenu.js
```diff
@@ -0,0 +1,174 @@
...
+      'slds-dropdown__list',
+    ];
+
+    const rest = omit(this.props, Object.keys(propTypes));
+
+    return (
+      <div className={cx(this.getClasses())} {...rest}>
+        {button}
+        {(isOpen || renderClosedDropdown) && <div className={cx(this.getDropdownClasses())}>
+          <ul className={cx(listClasses)} role="menu" onClick={this.handleItemClick}>
```
philippfrank: Why do we need the `onClick` at this point at all? The children `MenuItem` elements each have an `onClick`. Do I miss something?

brunottonurb: Reduce the number of event listeners. Apparently too many of those can slow down an app.

src/components/Menu/ControlledMenu.js
```diff
@@ -0,0 +1,174 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import omit from 'lodash/omit';
+import { getUniqueHash } from '../../utils';
+import MenuItem from './MenuItem';
+
+const propTypes = {
```
philippfrank: Why did you specify the `propTypes` outside of the class instead of `static propTypes = { ... }`; You could later reference them like `ControlledMenu.propTypes` if you need to.

src/components/Menu/ControlledMenu.js
```diff
@@ -0,0 +1,174 @@
...
+    const index = event.target.id.split('_')[2];
+    if (index && (index < React.Children.count(children))) {
+      React.Children.toArray(children)[index].props.onClick();
+    }
+    event.stopPropagation(); // stopPropagation so we can have nested menus
+  };
+
+  renderChildren = () => {
+    const { children, checkbox } = this.props;
+    if (!children) return null;
```
philippfrank: You have `children` required in the prop types, this is obsolete.

src/components/Menu/ControlledMenu.js
```diff
@@ -0,0 +1,174 @@
...
+
+      if (child.type !== MenuItem) {
+        return React.cloneElement(
+          child, { key: id }
+        );
+      }
+      return React.cloneElement(
+        child, {
+          selected: checkbox ? child.props.selected === true : null,
+          // with this we set the selected prop to true or false (no undef or null)
```
philippfrank: I assume you reference the line above in this comment. `true or false (no undef or null)` is wrong, you are actually setting it to null.

src/components/Menu/ControlledMenu.js
brunottonurb: I believe it's because we use [transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types).

src/components/Menu/ControlledMenu.js
fspoettel: @philippfrank The Menu re-work has two goals:

1. Improve the API, making it easier to use and extend in the future (SF has prototypes for sub menus now)
2. Speed up current features where a lot of Menus are used (MatchingConfiguration mounts ~25 menus with huge sub trees which slows rendering down significantly (seconds)

Binding one event listener here has the benefit that we do not have to loop over the child items and attach/detach event listeners when re-rendering. This can potentially be expensive when taking the new option `renderClosedDropdown` into account since we would have to rebind all event listeners every time the dropdown opens again. After implementing this, we can opt into performance optimizations in features such as MatchingConfig without negating the speed gains by binding/unbinding hundreds of EventListeners at once

src/components/Menu/ControlledMenu.js
fspoettel: @philippfrank @brunottonurb is right here. If you need a reference to the propType object, you need to persist it outside `Component.propTypes` since those will be stripped in production builds

src/components/Menu/ControlledMenu.js
philippfrank: @fspoettel Makes sense. Nevertheless -- unless I'm missing something -- the MenuItems have their `onClick`-handler. The items are just cloned. Did you intend to remove them at that point? https://github.com/propertybase/react-lds/pull/293/files/00d5f8748a161eec579e96d7e4481dde6d0d3317#diff-e75026ec5544dc1bd4f13c5484b4ee42R132

brunottonurb: Yes. I now set ´onClick` to `null` when cloning.

fspoettel: imo the best solution would be to attach an `onSelect` in `Menu` and call it with a hash`{ id, key }` of the MenuItem. This would free up `MenuItem` event listeners for custom logic. @brunottonurb unless I'm missing something after glancing over: How does nulling the `onClick` solve the problem when you still call it via `React.Children.toArray(children)[index].props.onClick();` in https://github.com/propertybase/react-lds/pull/293/files#diff-e75026ec5544dc1bd4f13c5484b4ee42R116?

brunottonurb: As far as I understand, the MenuItem `onClick` stays a prop, it's never rendered. So when rendering the MenuItems in ControlledMenu, I null the `onClick` prop, so it won't get rendered via `{...rest}` in MenuItem. ControlledMenu can still access the MenuItems `onClick`, because its props still contain the unchanged MenuItems.
I think this is more intuitive to use than an `onSelect`. When using the Menu, you don't even need to know we reduce the number of event listeners and pass your `onClick` to MenuItem like you'd always do.

philippfrank: Well, what @fspoettel was saying is that the *attachment/detachment* of a huge number of listeners on the menu item level is the problem. Since you clone the original children, you also clone the `onClick` prop which results in an event listener to be attached for each menu item. 

When I get the plan correctly here, we want only a handler on the entire menu itself (whether we call that `onSelect` or something else; maybe `onItemSelect`?) instead of one on each item.

This results in a click on an item will be caught by the event listener which was attached to the menu which then reads the items ID and passes that as an argument to the `onSelect` handler. 

5 menus on a page with each 10 menu items will then result in 5 event listeners instead of 50.

fspoettel: @brunottonurb 

> pass your onClick to MenuItem like you'd always do

There is nothing stopping you from doing that if we provide an `onSelect` callback. Attaching events to parent components is something you see a lot in bigger component libraries and it definitely makes sense when your component can contain hundreds of items and render frequently, making binding/unbinding events an expensive task. If we free up the event listener on `MenuItem` and let users set it on their own they can choose to bind `onClick` handlers to `MenuItem` and ignore that `onSelect` even exists. But having an `onSelect` and one instead of a couple hundred event listeners is something that should be possible.

fspoettel:
``` jsx
// Native Select in React
const select = (
  <select onChange={this.handle}>
    <option value="a">A</option>
    <option>B</option>
  </select>
);

// Suggested Menu API (Reference key is up for discussion, id would work, a data-attribute could be preferential. We could set that automatically when cloning the child)
const menu = (
  <Menu onSelect={this.handle}>
    <MenuItem id="a" data-value="a">A</MenuItem>
    <MenuItem id="b">b</MenuItem>
  </Menu>
);

// Still possible
const menuChildClick = (
  <Menu>
    <MenuItem onClick={this.handle}>A</MenuItem>
    <MenuItem onClick={this.handle}>B</MenuItem>
  </Menu>
```

src/components/Menu/MenuItem.js
```diff
@@ -61,9 +71,22 @@ const MenuDropdownListItem = (props) => {
...
+    >
+      <a
+        aria-checked={selected}
+        role={role}
+        data-value={dataValue}
+      >
+        <span
+          className="slds-truncate"
+          title={children}
+          data-value={dataValue}
```
fspoettel: Please remove this attribute here

src/components/Menu/MenuItem.js
```diff
@@ -61,9 +71,22 @@ const MenuDropdownListItem = (props) => {
...
+    <li
+      className={cx(sldsClasses)}
+      role="presentation"
+      onClick={onClick}
+      {...rest}
+    >
+      <a
+        aria-checked={selected}
+        role={role}
+        data-value={dataValue}
```
fspoettel: I don't get why this is added here and not on the li since you access dataset on li in ControlledMenu.

src/components/Menu/__tests__/Menu.spec.js
```diff
@@ -1,83 +1,37 @@
...
   });
 
-  it('uses nubbin class', () => {
-    mounted.setProps({ position: 'bottom-right', nubbin: true });
-    expect(mounted.find('div').last().hasClass('slds-nubbin_bottom-right')).toBeTruthy();
+  it('starts as closed by default and defaultOpens', () => {
+    let mounted = getComponent();
+    expect(mounted.find('div.slds-dropdown').length).toBe(0);
+    mounted = getComponent({ defaultOpen: true });
+    expect(mounted.find('div.slds-is-open').length).toBe(1);
```
fspoettel: All these length = 0|1 checked could be written like `.exists()).toBeTruthy()` `.exists()).toBeFalsy()`

src/components/Menu/__tests__/Menu.spec.js
```diff
@@ -1,83 +1,37 @@
...
-    mounted.setProps({ position: 'top-right' });
-    expect(mounted.find('div').last().hasClass('slds-dropdown_right')).toBeTruthy();
-  });
-
-  it('uses bottom class', () => {
-    mounted.setProps({ position: 'bottom-right' });
-    expect(mounted.find('div').last().hasClass('slds-dropdown_bottom')).toBeTruthy();
+    // expect(menu.hasClass('slds-is-open')).toBeFalsy();
+    expect(menu.find('button').length).toBe(1);
+    // menu.find('button').simulate('click');
```
fspoettel: Why are these commented out?

src/components/PageHeader/ObjectHome.js
```diff
@@ -4,7 +4,7 @@ import cx from 'classnames';
 import enhanceWithClickOutside from 'react-click-outside';
```
fspoettel: I think the Menu has been made optional for `ObjectHome` recently, please check whether this change affects your PR and merge the conflicts accordingly.

stories/Menu.js
```diff
@@ -1,84 +1,151 @@
...
-    onClick={action('custom Button clicked')}
+    onClick={action('Menu Button clicked')}
     aria-haspopup="true"
     className="slds-button_icon-border-filled"
   />
 );
 
 stories
-  .add('Default', () => (
+  .add('Controlled', () => (
```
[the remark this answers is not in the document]
brunottonurb: What do you mean with spacing provider?

stories/Menu.js
```diff
@@ -1,84 +1,151 @@
...
-        <MenuDropdownListItem
-          divider={boolean('Divider item 3', false)}
-        >Item 3</MenuDropdownListItem>
-        <MenuDropdownListItem
-          divider={boolean('Divider item 4', true)}
-        >Important last item</MenuDropdownListItem>
-      </MenuDropdownList>
+      <MenuItem
+        divider={boolean('Divider item 1', false)}
+        dataValue="item1"
```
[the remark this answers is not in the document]
brunottonurb: So I could remove it from propTypes?

src/components/Menu/MenuItem.js
brunottonurb: It does not work with `data-value` on the `li` for some reason. Then `data-value` is always undefined in the event.

fspoettel: Hm that is odd. I'm gonna take a look

src/components/Menu/ControlledMenu.js
```diff
@@ -0,0 +1,177 @@
...
+   */
+  nubbin: PropTypes.bool,
+  /*
+   * use onSelect instead of onClick on the MenuItems if you want reduce the number of event listeners
+   */
+  onSelect: PropTypes.func,
+  /**
+   * position relative to the menu button
+   */
+  position: PropTypes.oneOf(['top-left', 'top', 'top-right', 'bottom-left', 'bottom', 'bottom-right']),
```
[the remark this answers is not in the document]
brunottonurb: :ok_hand:

philippfrank:
Nice PR @brunottonurb

Issue #295, opened 2018-02-28
Minor: Picklist size prop comes with a warning in case full width is desired


PeterSiegl:
```
bugsnag.js?0123:661 Warning: Failed prop type: Invalid prop `size` of value `` supplied to `PicklistRaw`, expected one of ["small","medium","large"].
```

The sample prop given on the demo page is not available:
```
<PicklistRaw
  id="picklist-1"
  items={[{{key: "header1", label: "Header 1", isHeader: {true}}}, {{key: "option1", label: "Option 1", selected: {false}}}, {{key: "option2", label: "Option 2", selected: {false}}}, {{key: "option3", label: "Option 3", selected: {false}}}, {{key: "option4", label: "Option 4", selected: {false}}}, {{key: "option5", label: "Option 5", selected: {false}}}, {{key: "header2", label: "Header 2", isHeader: {true}}}, {{key: "option6", label: "Option 6", selected: {false}}}]}
  labelInput="Tis a input label"
  labelMultiselect="*multiselect label*"
  onSelect={selected()}
  placeholder="Tis a placeholder"
  size="text"
/>
```

Desired behavior: if the size prop is left out or null, the picklist should scale up to 100% width.

PR #297, opened 2018-03-13
Change Picklist default size prop, adapt story

brunottonurb:
Issue #295

### DIFF
--- a/stories/Form.js
+++ b/stories/Form.js
@@ -115,7 +115,7 @@ stories
       labelMultiselect={text('MultiSelect', '*multiselect label*')}
       onSelect={action('selected')}
       placeholder={text('Placeholder', 'Tis a placeholder')}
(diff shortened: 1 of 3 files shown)


stories/Form.js
```diff
@@ -115,7 +115,7 @@ stories
       labelMultiselect={text('MultiSelect', '*multiselect label*')}
       onSelect={action('selected')}
       placeholder={text('Placeholder', 'Tis a placeholder')}
-      size={select('Size', ['small', 'medium', 'large'], 'text')}
+      size={select('Size', ['', 'small', 'medium', 'large'], '') || undefined}
```
fspoettel: fyi there is a better way to do this once we upgrade `storybook`, see [here](https://github.com/storybooks/storybook/tree/master/addons/knobs#selectv2)

PR #299, opened 2018-03-15
Add option for auto-closing a picklist on select

extronics:


src/components/Form/Picklist.js
```diff
@@ -67,10 +75,10 @@ export class PicklistRaw extends Component {
   };
 
   renderPicklistItems() {
-    const { onSelect, items } = this.props;
+    const { items } = this.props;
 
     return items.map((item) => {
-      const boundClick = () => onSelect(item.key);
+      const boundClick = () => this.onSelect(item.key);
```
fspoettel: Mighty as well change method `onSelect` to `getSelectHandler(itemKey)` here as it avoids creating functions in render

src/components/Form/Picklist.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import enhanceWithClickOutside from 'react-click-outside';
+import { memoize } from 'lodash';
```
fspoettel: Importing from `lodash/memoize` is better here since `{} from 'lodash'` does not work with tree shaking for webpack<4 and requires `babel-plugin-lodash` to be present to reduce build size when requiring `react-lds`

PR #308, opened 2018-04-03
Menu rework

fspoettel:


brunottonurb:
This includes `Fix menu header markup for long titles` (slds-truncate class on header)

PR #311, opened 2018-04-10
Add custom outside click wrapper

PeterSiegl:


src/components/Datepicker/Datepicker.js
```diff
@@ -446,6 +445,16 @@ export class DatepickerRaw extends Component {
         )}
       </div>
     );
+
+    if (closeOnClickOutside) {
```
fspoettel: I'm not sure whether this is approach scales well. Would it make sense to add a `active` prop to `<ClickOutside />`? That way, we would avoid conditional render statements in every feature that uses this

PR #325, opened 2018-05-16
Refactor File, show titles on hover when no image set

fspoettel:


src/components/File/FileMedia.js
```diff
@@ -0,0 +1,50 @@
...
+};
+
+FileMedia.defaultProps = {
+  image: null,
+};
+
+FileMedia.propTypes = {
+  fileType: PropTypes.string.isRequired,
+  image: PropTypes.shape({
+    src: PropTypes.string,
```
PeterSiegl: order

PR #329, opened 2018-05-24
Release/7.0.1

PeterSiegl:


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 7.0.1
+
+* Small `Lookup` enhancements
```
fspoettel: Please be a bit more specific here ^^

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-lds",
-  "version": "7.0.0",
+  "version": "7.0.1",
```
fspoettel: This is not a bugfix so should get `7.1.0` version imo

CHANGELOG.md
PeterSiegl: kk

package.json
PeterSiegl: just some minor stuff but kk

fspoettel: https://semver.org/#spec-item-7 https://semver.org/#spec-item-6

PR #337, opened 2018-06-25
Enable tree-shaking

fspoettel:


.babelrc
```diff
@@ -28,10 +27,29 @@
...
+      ]
+    },
+    "commonjs": {
+      "presets": [
+        ["env", {
+          "targets": {
+            "browsers": ["last 2 versions"]
+          },
+          "modules": "commonjs",
+          "useBuiltIns": false
```
extronics: I think `false` is the default for `useBuiltIns`

PR #340, opened 2018-08-06
Tooltip: Add support for `overflow: hidden;` containers

quadreex:


src/components/Tooltip/ControlledTooltip.js
```diff
@@ -1,8 +1,11 @@
 import React from 'react';
```
brunottonurb: `import React, { PureComponent } from 'react';`

PR #347, opened 2018-09-07
Add SplitView component

quadreex:


src/components/SplitView/index.js
```diff
@@ -0,0 +1,117 @@
...
+
+	  if (isOpen) {
+	    onClose();
+	  } else {
+	    onOpen();
+	  }
+	}
+
+	renderMasterView() {
+    return this.state.isOpen
```
philippfrank: Maybe shorter: `this.state.isOpen && (...)` without ternary operator

src/components/SplitView/index.js
```diff
@@ -0,0 +1,117 @@
...
+	    masterMaxWidth,
+	    masterMinWidth,
+	  } = this.props;
+	  const { isOpen } = this.state;
+	  const masterStyle = isOpen
+	    ? { maxWidth: masterMaxWidth, minWidth: masterMinWidth }
+	    : null;
+
+	  return (
+  <div
```
philippfrank: Indentation seems to be wrong here

src/components/SplitView/index.js
```diff
@@ -0,0 +1,115 @@
+import React, { Component } from 'react';
```
fspoettel: It would make sense to split this into a `ControlledSplitView` and a light stateful wrapper `SplitView`. There are use cases where fully controlling the open state makes sense.

src/components/SplitView/index.js
```diff
@@ -0,0 +1,115 @@
...
+    masterMaxWidth: PropTypes.string,
+    masterMinWidth: PropTypes.string,
+    onOpen: PropTypes.func,
+    onClose: PropTypes.func,
+  };
+  static defaultProps = {
+    assistiveTextOpen: '',
+    assistiveTextClose: '',
+    className: null,
+    initialIsOpen: false,
```
fspoettel: Default should be `true` imho

src/components/SplitView/ToggleButton.js
```diff
@@ -0,0 +1,49 @@
...
+  const assistiveText = isOpen
+    ? assistiveTextOpen
+    : assistiveTextClose;
+  const buttonClassName = cx(
+    'slds-button slds-button_icon slds-split-view__toggle-button',
+    { 'slds-is-open': isOpen },
+  );
+
+  return (
+    <button
```
fspoettel: is it not possible to use an `IconButton` here? This seems to just reimplement a good portion of that component.

src/components/SplitView/__tests__/SplitView.spec.js
```diff
@@ -0,0 +1,39 @@
...
+  />
+);
+
+describe('<SplitView />', () => {
+  it('renders the correct markup for closed state', () => {
+    const mounted = shallow(renderSplitView());
+
+    expect(mounted.find('.slds-split-view_container').length).toBe(1);
+    expect(mounted.find('.slds-split-view_container').hasClass('slds-is-closed')).toBeTruthy();
+    expect(mounted.find('.slds-split-view').length).toBe(0);
```
fspoettel: Better to use `.exists` and a boolean check imo

src/components/SplitView/index.js
quadreex: It is strong antipattern to have 'true' default values for boolean props

quadreex: it breaks semantics of using them like normal html tags:
```
<SplitView initialIsOpen />
```

src/components/SplitView/ToggleButton.js
quadreex: Icon button brings slds styles, that break visuals.

src/components/SplitView/__tests__/SplitView.spec.js
quadreex: 👍

src/components/SplitView/index.js
fspoettel: I agree about it being an antipattern, `initialIsIsOpen` still seems like the default case though. Would it make sense to use `initialIsClosed` prop here instead then?

PR #349, opened 2018-09-09
Fix broken loading state for inputs, align markup with SLDS specs

fspoettel:


src/components/Form/InputRaw.js
```diff
@@ -99,22 +99,14 @@ const InputRaw = (props) => {
...
-
   const sldsClasses = [
     { 'slds-has-input-focus': isFocused },
     bare ? 'slds-input_bare' : 'slds-input',
     className
   ];
 
   return (
-    <span>
+    <Fragment>
```
quadreex: why not just use `<>` and `</>`

fspoettel: Last time I checked, babel did not support that syntax yet without a specific plugin which I could not get to work in Storybook. This might be outdated now though

fspoettel: @quadreex I don't think there is a way around this since we are still using Babel 6 in this project and JSX Fragment support starts with a late Babel 7 beta. I'd prefer to merge this for now and do the whole storybook / babel update at a later date when storybook support for babel 7 and webpack 4 is [out of beta](https://github.com/storybooks/storybook/issues/3805)

quadreex: 👍

PR #355, opened 2018-10-09
Release 7.2.4

tomlord90:


CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
 # Changelog
+## 7.2.4
+
+* Add support for `ComboboxDropdown` email modifier
+* Fixed: Removed blue border on the PDF template selector field
```
fspoettel: PDF template selector is not a react-lds component. We can use a technical formulation here: _Fixed: Opening Menus now does not trigger an incorrect ::focus state on parent_

PR #367, opened 2018-10-30
Allow grid components to have arbitrary DOM elements as containers

fspoettel:


src/components/Grid/__tests__/Column.spec.js
```diff
@@ -51,4 +51,9 @@ describe('<Column />', () => {
     mounted.setProps({ 'small-order': '3' });
     expect(mounted.find('.slds-col').hasClass('slds-small-order_3')).toBeTruthy();
   });
+
+  it('renders as arbitrary DOM node', () => {
```
philippfrank: `renders as footer node`

And to emphasize the fact that you really can render as arbitrary elements, you may use someting super weired like `marquee`. Unfortunately that's not part of HTML5, so maybe `wbr` 😈

fspoettel: `<blink>`

PR #368, opened 2018-11-05
Fix horizontal scroll for DataTable with fixed header

quadreex:


quadreex:
@fspoettel I haven't touched DatePicker component. Do you know, why the tests are broken?

fspoettel:
@quadreex Noticed that too, I'll investigate.

src/components/Table/Cell.js
```diff
@@ -98,7 +98,7 @@ const Cell = (props) => {
   };
 
   const wrapChildren = (content) => {
-    if (!!sortable && isHeader) {
+    if (isHeader) {
```
fspoettel: Shouldn't this return a span in case it is not a an action like `sortable`?

quadreex: It breaks visual styles for non-sortable fields

PR #385, opened 2018-12-07
Add prop type node to Slider, CheckboxToggle and ProgressBar label

PeterSiegl:


src/components/CheckboxToggle/CheckboxToggle.js
```diff
@@ -73,7 +73,10 @@ CheckboxToggle.propTypes = {
   /**
    * Checkbox label
    */
-  label: PropTypes.string.isRequired,
+  label: PropTypes.oneOfType([
+    PropTypes.string,
```
quadreex: If I am not mistaken `PropTypes.node.isRequired` can be used here directly, as it is compatible with strings also.
`PropTypes.node` - is the most generic type.

quadreex:
@PeterSiegl Could you please also change the type of `label` prop for `Pill` component?

src/components/CheckboxToggle/CheckboxToggle.js
```diff
@@ -73,7 +73,10 @@ CheckboxToggle.propTypes = {
...
+    PropTypes.node,
+  ]).isRequired,
```
fspoettel: Shouldn't `statusLabels` also be changed to keep this consistent?

fspoettel:
@quadreex Changing this for `Pill` is potentially hazardous to the expected behavior specified in the LDS. It relies on truncation classes that only work with text (and probably `inline`) content, anything else will break that behavior.

src/components/CheckboxToggle/CheckboxToggle.js
PeterSiegl: Agreed, I'll change other occurrences of this pattern too.

quadreex:
@fspoettel I already use inline components with Pill. And I get the warning. I think that decision should be made by the user of a component and not the component itself about allowing to use different labels.

fspoettel:
@quadreex I agree with this for the most part. We still should stick to the underlying LDS implementation as close as possible and [the docs](https://lightningdesignsystem.com/components/pills/) do not indicate support for anything but a string as pill content. May I ask what your use case is here? We may as well make an exception here and implement the propType if the usecase is reasonable

quadreex:
@fspoettel @PeterSiegl Example where I use markup inside Pill:
https://github.com/propertybase/pba-react/blob/1551e1979258a30cabdab27b95a0674ff7c1b134/app/DripCampaigns/components/InputField/components/ContactDateFieldInput.js#L42

fspoettel:
@quadreex Thanks! Let's make the change in that case.

PeterSiegl:
@quadreex @fspoettel I adjusted the label prop type for pills. can I merge this?

Issue #387, opened 2018-12-29
How is assetBasePath used?


yoDon:
The main README says "to use ReactLDS you will have to provide assetBasePath via the ReactContext" but it provides no guidance on what assetBasePath must be set to (the snippet in the README sets it to the empty string, which does not actually seem to be required despite the stern language that we must provide it).

What is assetBasePath used for, when should it be set, and what should it be set to?

brunottonurb:
The `assetBasePath` should point to the folder where the salesforce lightning-design-system fonts, icons, images, and styles are located. This could be `assets/`, depending on your setup.
The example seems broken, the naming is also inconsistent. I will look into it.

fspoettel:
@yoDon The `assetBasePath` is required to use `Icon` and `Avatar` components. Those work by using a svg/img reference from the LDS icon sprites via `xlink:href`. We have to provide the path to construct those URLs since we have no knowledge how the LDS is hosted in a given project.

If you are using `react-lds` outside of a Salesforce org, @brunottonurb's instructions are accurate. Inside of a Salesforce org, `assetBasePath` should be set to `{!JSENCODE(URLFOR($Asset.SLDS, '/'))}` from the surrounding LightningComponent / Visualforce page.

PR #388, opened 2019-01-08
Clarify Context in README

brunottonurb:
https://github.com/propertybase/react-lds/issues/387

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,21 +50,26 @@ import PropTypes from 'prop-types';
 class AssetPathProvider extends Component {
   getChildContext() {
     return {
-      assetBasePath: '',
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,21 +50,26 @@ import PropTypes from 'prop-types';
...
+      assetBasePath: 'assets/',
     };
   }
 
   render() {
     const { children } = this.props;
-    return Children.only(children);
+
+    return (
+      <div>
```
fspoettel: Why change this?

fspoettel: In the current state, a `div` wrapper is implied which is also not ideal but may be less confusing

PR #396, opened 2019-02-08
Combobox improvements

PeterSiegl:


src/components/Form/InputRaw.js
```diff
@@ -105,7 +105,7 @@ const InputRaw = React.forwardRef((props, ref) => {
     <Fragment>
       {renderIconLeft()}
       <input
-        {...omit(rest, ['label'])}
+        {...omit(rest, ['isFocused', 'label'])}
```
fspoettel: This is not a fix, `isFocused` should not be set on the component in the first place. If you want to handle focus, supply a `renderInput` function and bind a ref to it

PeterSiegl: It was set in the `Lookup` component, removed the prop there and revert the above change.

PR #397, opened 2019-02-11
Add custom required & error handling to Combobox

quadreex:


src/components/Combobox/components/ComboboxCore.js
```diff
@@ -18,6 +18,10 @@ class ComboboxCore extends Component {
      * Classname that will be applied to the <FormElement> wrapping input and listbox
      */
     comboboxClassName: PropTypes.string,
+    /**
```
fspoettel: Can you add a note to all three props and note that these are custom additions to the LDS combobox specs?

src/components/Combobox/components/ComboboxDropdown.js
```diff
@@ -75,6 +79,7 @@ const ComboboxDropdown = React.forwardRef((props, ref) => {
         </div>
       </FormElementControl>
       {renderListbox && renderListbox()}
+      {!isErrorHidden && <FormElementError error={error} id={`error-${id}`} />}
```
fspoettel: I would argue that we should strip these props in everything but `Picklist` components. Otherwise, beneath the listbox seems like an odd place to render these

src/components/Combobox/components/ComboboxCore.js
quadreex: 👍

PR #404, opened 2019-02-15
Release 7.4.0

fspoettel:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-lds",
-  "version": "7.3.2",
+  "version": "7.3.3",
```
quadreex: an inconsistency of versions in `package.json` and `CHANGELOG.md`

fspoettel: fixed already

PR #406, opened 2019-02-24
Add SetupAssistant

philippfrank:


philippfrank:
@fspoettel I would appreciate some early feedback, my React is a little rusty. 

For this to make sense I would next implement the (vertical & horizontal) progress component (but in a separate PR). Also the `Card` component would need the possibility to render a pure header / no content to work as in the [example](https://www.lightningdesignsystem.com/components/setup-assistant/#In-a-Card).

src/components/Progress/ProgressRing.js
```diff
@@ -75,7 +76,7 @@ const ProgressRing = (props) => {
       >
         <svg viewBox="-1 -1 2 2">{getPath(percComplete / 100, isForceComplete, variant)}</svg>
       </div>
-      <div className="slds-progress-ring__content">{iconEl}</div>
+      <div className="slds-progress-ring__content">{content}</div>
```
philippfrank: The setup assistant uses text content within the progress nodes. As that can basically be anything, I added the possibility to add `children`.

src/components/SetupAssistant/SetupAssistantItem.js
```diff
@@ -0,0 +1,137 @@
...
+SetupAssistantItem.defaultProps = {
+  children: null,
+  stepProgress: null,
+  step: null,
+  isExpanded: false,
+  onExpand: null,
+  renderAddon: () => null,
+  renderExpanded: () => null,
+  // eslint-disable-next-line react/prop-types
+  renderProgress: ({ stepProgress, step }) => (
```
philippfrank: Can the PropType warnings be disabled for default `func` props that take arguments?

src/components/Progress/ProgressRing.js
```diff
@@ -106,6 +108,7 @@ ProgressRing.propTypes = {
     warning: PropTypes.string.isRequired,
     complete: PropTypes.string.isRequired,
   }),
+  children: PropTypes.node,
```
fspoettel: It would make sense to document that children takes precedence over status / customIcon

src/components/SetupAssistant/SetupAssistantItem.js
```diff
@@ -0,0 +1,137 @@
...
+                  <div className="slds-media__figure">
+                    {renderProgress({ stepProgress, step })}
+                  </div>
+                )}
+                <div className="slds-media__body slds-m-top_x-small">
+                  <div className="slds-media">
+                    <div className="slds-setup-assistant__step-summary-content slds-media__body">
+                      <h3 className="slds-setup-assistant__step-summary-title slds-text-heading_small">
+                        {onExpand ? (
+                          <button
```
fspoettel: Can you use a `<Button />` here?

src/components/SetupAssistant/SetupAssistantItem.js
```diff
@@ -0,0 +1,137 @@
...
+  renderProgress,
+  renderExpanded,
+  renderAddon,
+  onExpand
+}, { assetBasePath }) => (
+  <li className="slds-setup-assistant__item">
+    <article className="slds-setup-assistant__step">
+      <div className={cx('slds-summary-detail', { 'slds-is-open': isExpanded })}>
+        {onExpand && (
+          <button
```
fspoettel: I'm pretty sure that you could use an `<IconButton />` here, which would also alleviate the need for passing `contextTypes`. See the `With Custom ButtonIcon` example

src/components/SetupAssistant/SetupAssistantItem.js
```diff
@@ -0,0 +1,137 @@
...
+          >
+            <svg className="slds-button__icon slds-summary-detail__action-icon">
+              <use xlinkHref={`${assetBasePath}icons/utility-sprite/svg/symbols.svg#switch`} />
+            </svg>
+          </button>
+        )}
+        <div className="slds-container_fluid">
+          <div className="slds-summary-detail__title">
+            <div className="slds-setup-assistant__step-summary">
+              <div className="slds-media">
```
fspoettel: You should be able to use a `<MediaObject />` here if you add a `bodyClassName` prop to it.

src/components/SetupAssistant/SetupAssistantItem.js
```diff
@@ -0,0 +1,137 @@
...
+            type="button"
+            className="slds-button slds-button_icon slds-m-right_x-small slds-m-top_x-small"
+            onClick={onExpand}
+          >
+            <svg className="slds-button__icon slds-summary-detail__action-icon">
+              <use xlinkHref={`${assetBasePath}icons/utility-sprite/svg/symbols.svg#switch`} />
+            </svg>
+          </button>
+        )}
+        <div className="slds-container_fluid">
```
fspoettel: You can use `<Container flavor="fluid>` here

src/components/SetupAssistant/SetupAssistantItem.js
```diff
@@ -0,0 +1,137 @@
...
+          <div className="slds-summary-detail__title">
+            <div className="slds-setup-assistant__step-summary">
+              <div className="slds-media">
+                {stepProgress && (
+                  <div className="slds-media__figure">
+                    {renderProgress({ stepProgress, step })}
+                  </div>
+                )}
+                <div className="slds-media__body slds-m-top_x-small">
+                  <div className="slds-media">
```
fspoettel: You could also use a `<MediaObject />` here

src/components/SetupAssistant/SetupAssistantItem.js
fspoettel: Nope, unfortunately not.

src/components/SetupAssistant/SetupAssistantItem.js
philippfrank: Wouldn't it be nicer to add a prop `iconClassName` to `IconButton` and pass that through?

fspoettel: Hm I'm not sure about that. This is a fringe use case that should only be encountered when implementing other LDS components and I don't think we should litter `IconButton` with a prop in that case

src/components/Button/Base/Button.js
```diff
@@ -90,20 +90,17 @@ Button.propTypes = {
    * Optional additional className
    */
   className: PropTypes.string,
-  /**
```
fspoettel: We have these because `Storybook` did not show enums at some point. Would make sense to remove all of these since Storybook4 seems to support it?

stories/Button.js
```diff
@@ -23,6 +23,7 @@ const buttonFlavors = [
   'success',
   'text-destructive',
   'stretch',
+  'reset',
```
fspoettel: Not sure if we should add this since it is technically not a flavor. From what I understand about this, using `className` seems more approriate as it again is not a part of the "public" component API

PR #409, opened 2019-03-03
Add SetupAssistant, SummaryDetail

philippfrank:


stories/SummaryDetail.js
```diff
@@ -0,0 +1,26 @@
...
+import { SummaryDetail } from '../src';
+import './utils/global.css';
+
+const stories = storiesOf('Summary Detail', module);
+
+stories
+  .add('Base', () => (
+    <div>
+      <SummaryDetail
+        renderOpenContent={() => <div>Summary Content</div>}
```
fspoettel: From an API perspective, it would probably make more sense to pass the components' content as a node, maybe even as `children`. A function prop feels a bit clunky here. I would expect `title` to be it's own prop since I do not want to know about the implementation details (h3 and classes). `children` IMO is the wrong place for the title. I have two alternative suggestions:

``` jsx
// renderTitle is optional
<SummaryDetail title="foo" renderTitle={title => title}>
  {({ isOpen }) => 'foo')}
</SummaryDetail>
```

``` jsx
// Children is only rendered when `isOpen`
<SummaryDetail title="foo" renderTitle={title => title}>
  foo
</SummaryDetail>
```

philippfrank: You are right. Suggestion 1) is nice, it will prevent closed content from being rendered

src/components/SetupAssistant/SetupAssistantItem.js
```diff
@@ -0,0 +1,123 @@
...
+  );
+
+  return (
+    <li className="slds-setup-assistant__item">
+      <article className="slds-setup-assistant__step">
+        <SummaryDetail
+          isOpen={isOpen}
+          onOpen={onOpen}
+          renderTitle={renderTitle}
+          iconButtonClassName="slds-m-top_x-small"
```
philippfrank: @fspoettel Why this: The setup assistant item requires an additional class on the expand icon button

src/components/SetupAssistant/SetupAssistantItem.js
```diff
@@ -0,0 +1,131 @@
...
+        >
+          {renderOpenContent && (() => (
+            <div className="slds-setup-assistant__step-detail">
+              {renderOpenContent()}
```
fspoettel: Correct me if I'm wrong, but shouldn't this receive `{ isOpen }` from parent closure now too?

PR #410, opened 2019-03-03
Add ProgressIndicator

philippfrank:


src/components/Progress/Progress.js
```diff
@@ -0,0 +1,84 @@
...
+  return (
+    <div>
+      <div
+        className={cx('slds-progress', {
+          [`slds-progress_${flavor}`]: flavor,
+          'slds-progress_vertical': isVertical,
+        })}
+      >
+        <ol className="slds-progress__list">
+          {React.Children.map(children, child => React.cloneElement(child, { isVertical }))}
```
philippfrank: @fspoettel What's the cleanest way to pass the `isVertical` prop to each child?

fspoettel: This is the standard way to do it. You could go the extra mile to support `Fragments` (see `ButtonGroup` for an example for how to do it)

fspoettel: Since there is other logic depending on the child count though, I'd say we better disallow supplying `Fragments`, what do you think?

philippfrank: Would you enforce that via a custom PropType that checks on the children type and/or filter out "invalid" children?

fspoettel: I'd just document it for now

src/components/Progress/ProgressItem.js
```diff
@@ -0,0 +1,91 @@
...
+import { Icon } from '../Icon';
+
+const ProgressItem = ({
+  assistiveText,
+  isActive,
+  isError,
+  isCompleted,
+  isVertical,
+  children,
+  isSuccess,
```
fspoettel: By convention, we spread the `...rest` props to public components (`className` would have to be merged into cx). You can ignore this here if we decide to implement an array prop as explained above.

src/components/Progress/ProgressItem.js
```diff
@@ -0,0 +1,91 @@
...
+          {children || assistiveText}
+        </div>
+      </li>
+    );
+  }
+
+  return (
+    <li className={className}>
+      {!isCompleted && !isError && !isSuccess ? (
+        <button
```
fspoettel: Please use `Button`

src/components/Progress/ProgressItem.js
```diff
@@ -0,0 +1,91 @@
...
+        <div className="slds-progress__item_content slds-grid slds-grid_align-spread">
+          {children || assistiveText}
+        </div>
+      </li>
+    );
+  }
+
+  return (
+    <li className={className}>
+      {!isCompleted && !isError && !isSuccess ? (
```
fspoettel: This condition is the exact same above, it would make sense to extract it

src/components/Progress/Progress.js
```diff
@@ -0,0 +1,101 @@
...
+Progress.propTypes = {
+  /**
+   * Assistive text for the progress bar
+   */
+  assistiveText: PropTypes.string,
+  /**
+   * Progress steps. Need to be ProgressItem components
+   */
+  children: PropTypes.node,
+  /**
```
fspoettel: Bigger picture: I think we should consider two things here:

1. Split `<ProgressItem isVertical />` into `<VerticalProgressItem />` for the reason that `children` is only valid in one and not the other. I can see it being confusing.
2. We could think about accepting an array of progress objects instead of children:

``` jsx
[{ 
  ...,
  assistiveText: 'foo',
  content: (<p>Foo</p>)
}]
```

This has one big advantage: The public API for vertical lists would only be `isVertical`, the rest would be up to the component itself to implement.

src/components/Progress/Progress.js
```diff
@@ -0,0 +1,101 @@
...
+  progress,
+  progressLabel,
+}) => {
+  const progressPercent = calculateProgressPercent(
+    progress,
+    React.Children.count(children)
+  );
+
+  return (
+    <div>
```
fspoettel: Intentional wrapper div?

src/components/Progress/Progress.js
```diff
@@ -0,0 +1,101 @@
...
+      <div
+        className={cx('slds-progress', {
+          [`slds-progress_${flavor}`]: flavor,
+          'slds-progress_vertical': isVertical,
+        })}
```
fspoettel: As stated below, please allow spreading `...rest` here

src/components/Progress/ProgressItem.js
```diff
@@ -0,0 +1,91 @@
...
+  isActive: false,
+  isCompleted: false,
+  isError: false,
+  assistiveText: null,
+  isVertical: false,
+  children: null,
+  isSuccess: false,
+};
+
+ProgressItem.propTypes = {
```
fspoettel: Sort.

Also: Should have explanations. You can ignore this here if we decide to implement an array prop as explained above.

src/components/Progress/Progress.js
philippfrank: Nope, leftover of the popup as SLDS implemented it. Can be removed for now. 👍

PR #412, opened 2019-03-04
Add ContextBar

philippfrank:


fspoettel:
@philippfrank To answer your questions:

1. We should go with the documentation section labels, classes sometimes are shared across sections (e.g. combobox)
2. Passing a `renderLink({ children, href })` function would be sufficient. For `react-router`, it would look like:
``` jsx
  <ContextBar
    renderLink={({ children, className, href }) => (
      <Link className={className} to={href}>{children}</Link>
    )}
```

PR #414, opened 2019-03-22
Upgrade to Storybook^5

fspoettel:


fspoettel:
@PeterSiegl will squash it and fix

PR #415, opened 2019-03-22
Implement expandable listboxes for combobox results

fspoettel:
Im

### DIFF
--- a/src/components/Combobox/components/ComboboxGroupedListbox.js
+++ b/src/components/Combobox/components/ComboboxGroupedListbox.js
@@ -1,42 +1,141 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import cx from 'classnames';
 import { Icon } from '../../Icon';
+import { Button } from '../../Button';
 import { Listbox, Pill } from '../../Pill';
+import { withThrottle } from '../utils/withThrottle';
 
-export class ComboboxGroupedListbox extends Component {
+const isPropChanged = prop => (a, b) => a[prop] !== b[prop];
+
+const isLengthChanged = isPropChanged('length');
+const isIdChanged = isPropChanged('id');
+const isOpenChanged = isPropChanged('isOpen');
+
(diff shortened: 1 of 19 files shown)


src/components/Combobox/components/ComboboxGroupedListbox.js
```diff
@@ -0,0 +1,222 @@
...
+      >
+        {({ overflowCount }) => (
+          <div className={cx(sldsClasses)}>
+            {!isExpanded && overflowCount > 0 && (
+              <span className="slds-listbox-toggle" aria-hidden="true">
+                <Button
+                  tabIndex="-1"
+                  flavor="none"
+                  onClick={onExpand}
+                  title={`+${overflowCount} more`}
```
quadreex: shouldn't we support custom labels via props?

src/components/Combobox/utils/withThrottle.js
```diff
@@ -0,0 +1,20 @@
...
+  const throttled = (...args) => {
+    if (currentExecution == null) {
+      currentExecution = requestAnimationFrame(later(this, args));
+    }
+  };
+
+  return throttled;
+};
+
+export const withThrottle = Cmp => props => <Cmp {...props} throttleFn={rafThrottle} />;
```
quadreex: `cmp` is commonly used shortcut for `compare`. IMHO better would be `Component` or `Comp`

src/components/Combobox/utils/withThrottle.js
```diff
@@ -0,0 +1,20 @@
...
+  let currentExecution;
+
+  const later = (context, args) => () => {
+    currentExecution = null;
+    cb.apply(context, args);
+  };
+
+  const throttled = (...args) => {
+    if (currentExecution == null) {
+      currentExecution = requestAnimationFrame(later(this, args));
```
quadreex: have you tested that `this` works properly with arrow functions here?

PR #435, opened 2019-06-21
Grouped Combobox

fspoettel:


src/components/GroupedCombobox/__tests__/GroupedCombobox.spec.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import GroupedCombobox from '../GroupedCombobox';
+
+const getComponent = (props = {}) => shallow(
```
fspoettel: TODO: Remove

src/components/Combobox/components/ComboboxDropdown.js
```diff
@@ -55,6 +58,32 @@ const ComboboxDropdown = React.forwardRef((props, ref) => {
...
+            id={listboxId}
+            ref={ref}
+            role="listbox"
+            tabIndex="-1"
+          >
+            {children}
+          </div>
+        </div>
+      </div>
+    </Fragment>
```
quadreex: Why do we need `Fragment` here?

src/components/GroupedCombobox/GroupedComboboxContainer.js
```diff
@@ -0,0 +1,69 @@
...
+    const {
+      id,
+      isOpen,
+      items,
+      onToggle,
+      selectedItems
+    } = this.props;
+
+    return (
+      <Picklist
```
quadreex: Haven't we deprecated `Picklist` component?

src/components/Combobox/components/ComboboxDropdown.js
fspoettel: Good find. We actually don't, it's a leftover from an earlier iteration. Will remove

src/components/GroupedCombobox/GroupedComboboxContainer.js
fspoettel: Yes and no.`Picklist` currently is a reexport of `BaseCombobox` as intended in the LDS. `Picklist` seems to be the official name of the component. `Picklist` was its own beast back before `Comboboxes` where a thing in the LDS and _that_ impl. has been removed from this project with the last major version

quadreex: Shouldn't we just use it here as `BaseCombobox`? IMHO it will be more clean

fspoettel: Yes, I will change it

PR #437, opened 2019-06-28
Fix child rendering in ControlledMenu

PeterSiegl:
@fspoettel I stumbled upon an inconsistency when rendering child components inside a `Menu`. Locally, condition https://github.com/propertybase/react-lds/compare/fix-controlled-menu?expand=1#diff-e75026ec5544dc1bd4f13c5484b4ee42L113 (`ƒ MenuItem(props) {...}`) is truthy and the selected state is applied correctly but in the live version https://github.com/propertybase/react-lds/compare/fix-controlled-menu?expand=1#diff-e75026ec5544dc1bd4f13c5484b4ee42L119 (`ƒ (e){...}`) gets executed and requires the `checkbox` prop. Can we get rid of that inconsistent check or is there a reason I'm not seeing?

### DIFF
--- a/src/components/Menu/ControlledMenu.js
+++ b/src/components/Menu/ControlledMenu.js
@@ -11,10 +11,6 @@ const propTypes = {
    */
   button: PropTypes.element.isRequired,
   /**
-   * set to true if menuitems should be menuitemcheckboxes
-   */
-  checkbox: PropTypes.bool,
-  /**
    * should be MenuItems or MenuSubHeaders
    */
   children: PropTypes.node.isRequired,
@@ -81,7 +77,6 @@ class ControlledMenu extends Component {
 
   static defaultProps = {
     className: null,
-    checkbox: false,
     height: null,
     heightIcon: null,
     isOpen: false,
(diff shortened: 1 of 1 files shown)


src/components/Menu/ControlledMenu.js
```diff
@@ -104,24 +99,13 @@ class ControlledMenu extends Component {
...
-        });
-      }
-
-      return React.cloneElement(
-        child, {
-          selected: checkbox ? child.props.selected === true : null,
-          key: id,
-        }
-      );
+      return React.cloneElement(child, { key: id });
```
fspoettel: IIRC there is a reason for this check - I will look into it tomorrow. Not sure what this accomplishes and why it's failing.

fspoettel: Update: This check was apparently added to distinguish between different child accessibility roles. It abuses the boolean `selected` by assigning a semantic meaning to the value `null`. The assignment `checkbox ? child.props.selected === true : null` accomplishes an explicit cast to a bool. value when the `checkbox` prop is truthy, otherwise it explicitely casts to `null`.

@PeterSiegl This behavior is correct from an accessibility standpoint as far as I can tell from looking at the LDS specification. The implementation seems hacky. Does this cause an unrelated bug for you?

PeterSiegl: @fspoettel it's not really a bug the different behavior between production and localhost is just confusing. for production I have to set the `checkbox` prop which is not really obvious since locally everything works fine.

fspoettel:
Closing this one for now - feel free to reopen if you decide to clean up the underlying component logic as discussed

PR #444, opened 2019-07-30
Implement VisualPicker

fspoettel:


fspoettel:
I will implement it as it seems useful. It's a bit different from the others in both layout and structure, I wonder why SF included it in this particular blueprint :thinking:

PR #445, opened 2019-07-30
Implement ScopedNotifications

fspoettel:


quadreex:
Looks good. 
Is it possible to add support for `warning` theme?

fspoettel:
No, it does not support "real" themes right now unfortunately :(

PR #449, opened 2019-09-05
Implement FileSelector

fspoettel:


src/components/FileSelector/FileSelector.js
```diff
@@ -0,0 +1,170 @@
...
+  fatalError,
+  label,
+  multiple,
+  onDragEnter,
+  onDragLeave,
+  onDrop,
+  selectorText,
+  selectorType,
+}) => {
+  const errorStr = fatalError || error;
```
quadreex: What use cases we have for non-fatal error?

quadreex: Why not just expose `isDisabled` prop? IMHO it will be more universal. We may have situations where we want to disable FileSelector without any errors

fspoettel: Non-fatal error: An uploaded file was rejected but we can accept other uploads
Fatal error: The file selector is not usable due to missing permissions etc.

Salesforce does not expose `disabling` without an error string according to the docs, I don't think that we should break that expectation

PR #457, opened 2019-11-06
Add chat component

PeterSiegl:
Implements https://lightningdesignsystem.com/components/chat

### DIFF
--- a/src/components/Chat/Chat.js
+++ b/src/components/Chat/Chat.js
@@ -1,28 +1,35 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { get } from 'lodash-es';
+import { get, uniqueId } from 'lodash-es';
 import cx from 'classnames';
 
-import { uniqueId } from '../../utils';
 import { CHAT_ITEM_TYPES } from './utils';
 import ChatListItem from './ChatListItem';
 
-const Chat = ({ isPastChat, items }) => {
+const Chat = ({
+  className,
+  isPastChat,
+  items,
+}) => {
   const sectionClassNames = cx([
     'slds-chat',
     { 'slds-chat_past': isPastChat },
+    className,
   ]);
 
   return (
     <section role="log" className={sectionClassNames}>
       <ul className="slds-chat-list">
         {items.map((message, i) => {
-          const { author: messageAuthor, messageType } = message;
+          const {
+            author: messageAuthor,
+            type: messageType,
+          } = message;
           let isConsecutive = false;
           let isFirstMessage = false;
 
           const { author: prevAuthor, isStart: prevIsStart } = get(items, [i - 1], {});
-          const { author: nextAuthor, messageType: nextType } = get(items, [i + 1], {});
+          const { author: nextAuthor, type: nextType } = get(items, [i + 1], {});
 
           if (!isPastChat && messageType === nextType && messageAuthor === nextAuthor) {
             isConsecutive = true;
@@ -30,13 +37,15 @@ const Chat = ({ isPastChat, items }) => {
 
           isFirstMessage = prevIsStart || prevAuthor !== messageAuthor;
 
-          const otherProps = {
-            isConsecutive,
-            isFirstMessage,
-            isPastChat,
-          };
-
-          return <ChatListItem key={uniqueId(i, 'chat-list-item-')} {...message} {...otherProps} />;
+          return (
+            <ChatListItem
+              isConsecutive={isConsecutive}
+              isFirstMessage={isFirstMessage}
+              isPastChat={isPastChat}
+              key={uniqueId('chat-list-item-')}
+              {...message}
+            />
+          );
         })}
       </ul>
     </section>
@@ -44,10 +53,12 @@ const Chat = ({ isPastChat, items }) => {
 };
 
 Chat.defaultProps = {
+  className: null,
   isPastChat: false,
 };
 
 Chat.propTypes = {
+  className: PropTypes.string,
   isPastChat: PropTypes.bool,
   items: PropTypes.arrayOf(PropTypes.shape({
     messageType: PropTypes.oneOf(Object.values(CHAT_ITEM_TYPES)),
--- a/src/components/Chat/Chat.js
+++ b/src/components/Chat/Chat.js
@@ -1,15 +1,22 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { get, uniqueId } from 'lodash-es';
+import { get } from 'lodash-es';
 import cx from 'classnames';
 
 import { CHAT_ITEM_TYPES } from './utils';
 import ChatListItem from './ChatListItem';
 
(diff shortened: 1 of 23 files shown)


src/components/Chat/Chat.js
```diff
@@ -0,0 +1,65 @@
...
+          const { author: prevAuthor, isStart: prevIsStart } = get(items, [i - 1], {});
+          const { author: nextAuthor, messageType: nextType } = get(items, [i + 1], {});
+
+          if (!isPastChat && messageType === nextType && messageAuthor === nextAuthor) {
+            isConsecutive = true;
+          }
+
+          isFirstMessage = prevIsStart || prevAuthor !== messageAuthor;
+
+          const otherProps = {
```
fspoettel: Can spread these to the item directly

src/components/Chat/Chat.js
```diff
@@ -0,0 +1,65 @@
...
+            isConsecutive,
+            isFirstMessage,
+            isPastChat,
+          };
+
+          return <ChatListItem key={uniqueId(i, 'chat-list-item-')} {...message} {...otherProps} />;
```
fspoettel: This is the same as using an `index` and will result in a re-render on every update. What about using `timestamp+type` as identifier?

src/components/Chat/Chat.js
```diff
@@ -0,0 +1,65 @@
...
+  items: PropTypes.arrayOf(PropTypes.shape({
+    messageType: PropTypes.oneOf(Object.values(CHAT_ITEM_TYPES)),
+    attachment: PropTypes.shape({
+      fileType: PropTypes.string.isRequired,
+      isLink: PropTypes.bool,
+      meta: PropTypes.string,
+      src: PropTypes.string,
+      title: PropTypes.string.isRequired,
+    }),
+    message: PropTypes.node,
```
fspoettel: `.node` does not really make sense here, ideally `.shape()` can be used

src/components/Chat/ChatListItem.js
```diff
@@ -0,0 +1,59 @@
...
+
+    return 'inbound';
+  }
+
+  const classNamesItem = cx([
+    'slds-chat-listitem',
+    { [`slds-chat-listitem_${getListItemCssClass()}`]: isBookend || !isPastChat },
+  ]);
+
+  const renderListItem = () => {
```
fspoettel: Maybe inline this function in render?

```jsx
<li className={classNamesItem}>
  {() => {
    if (isEvent) return <ChatListItemEvent {...props} isError={isError} />;
    if (isBookend) return <ChatListItemBookend {...props} isEnd={isEnd} />;
    return <ChatListItemMessage {...props} />;
  }()}
</li>
```

src/components/Chat/ChatListItemBookend.js
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <div className={classNamesBookend}>
+      <Icon
+        className="slds-chat-icon"
+        icon={isEnd ? 'end_chat' : 'chat'}
+        size="x-small"
+        sprite="utility"
+        svgClassName="slds-icon-text-default"
+      />
+      {renderMeta(message, timestamp)}
```
fspoettel: Why is this a function istd. of a React component?

src/components/Chat/ChatListItemMessage.js
```diff
@@ -0,0 +1,219 @@
...
+            sprite="utility"
+          />
+          {labelResend}
+        </Button>
+      </Grid>
+    )
+    : !isConsecutive && renderMessageMeta(author, timestamp)
+  );
+
+  const renderMessageBody = () => {
```
fspoettel: Should be a React component rather than a function

src/components/Chat/ChatListItemMessage.js
```diff
@@ -0,0 +1,219 @@
...
+
+    return (
+      <div className="slds-chat-message__meta" aria-label={`${meta} ${timestamp}`}>
+        {renderedAuthor} &bull; {timestamp}
+      </div>
+    );
+  };
+  /* eslint-enable react/jsx-one-expression-per-line */
+
+  const renderMeta = () => (isDeliveryFailure
```
fspoettel: Should be a React component that also inlines `renderMessageMeta`

src/components/Chat/ChatListItemMessage.js
```diff
@@ -0,0 +1,219 @@
...
+        icon={isUnsupportedType ? 'warning' : 'error'}
+        svgClassName={classNamesStatusIcon}
+        size="x-small"
+        sprite="utility"
+      />
+      <span>{msg}</span>
+    </React.Fragment>
+  );
+
+  const renderFile = () => {
```
fspoettel: Should be a React component

src/components/Chat/ChatListItemMessage.js
```diff
@@ -0,0 +1,219 @@
...
+    { [`slds-chat-message__image_${getMessageSourceCssClass()}`]: !isLink && isImage },
+    { 'slds-chat-message__file': hasAttachment && !isLink && !isImage },
+    { [`slds-chat-message__file_${getMessageSourceCssClass()}`]: hasAttachment && !isLink && !isImage },
+  ]);
+  const classNamesStatusIcon = cx([
+    { 'slds-icon-text-warning': isUnsupportedType },
+    { 'slds-icon-text-error': isDeliveryFailure },
+  ]);
+
+  const renderStatusMessage = msg => (
```
fspoettel: Should be a React component

src/components/Chat/Chat.js
PeterSiegl: Why shape? A message can only contain a `string` or a `node`.

src/components/Chat/Chat.js
PeterSiegl: The timestamp is not really unique... Usually it's only on a per minute base.

fspoettel: Would it make sense to add an `id` to the data model then? As it is, using `uniqueId` here in render does not have the desired effect and is misleading

src/components/Chat/Chat.js
fspoettel: Sorry, I misread the render code `item.map(message => {})`. Can you rename `message` in the loop to `item` to help clear this up?

PeterSiegl: Sure 👍

src/components/Chat/Chat.js
PeterSiegl: Yep, there is nothing else that qualifies as a key.

src/components/Chat/ChatListItem.js
fspoettel: This was marked as resolved, can you pinpoint the reason why you chose to ignore this comment?

PeterSiegl: Just forgot to make it self-invoking. Done!

PR #458, opened 2019-11-15
Implement Tile & NameValueList

fspoettel:


src/components/NameValueList/NameValueList.js
```diff
@@ -0,0 +1,124 @@
...
+  ...rest
+}) => (
+  <El
+    {...rest}
+    className={cx([className, 'slds-truncate'])}
+    title={title || (typeof children === 'string' ? children : '')}
+  >
+    {children}
+  </El>
+
```
philippfrank: Remove empty line

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+* Added: `NameValueList.{List,Name,Value}` utility components
+* Added: `Tile` component
```
quadreex: please add Chat component here

src/components/NameValueList/NameValueList.js
```diff
@@ -0,0 +1,123 @@
...
+  </El>
+);
+
+NameOrValue.defaultProps = {
+  className: null,
+  title: null,
+};
+
+NameOrValue.propTypes = {
+  as: PropTypes.string.isRequired,
```
quadreex: Doesn't SF have a limitation on which lags can be used?

src/components/Tile/Tile.js
```diff
@@ -0,0 +1,106 @@
...
+
+  const tileClasses = cx([
+    'slds-tile',
+    { 'slds-tile_board': board },
+    { 'slds-hint-parent': !!action },
+    className,
+  ]);
+
+  const tileContent = (
+    <React.Fragment>
```
quadreex: Do we have plans to finally upgrade our deps? To get rid of `<React.Fragment>`

src/components/NameValueList/NameValueList.js
fspoettel: No they dont

src/components/Tile/Tile.js
fspoettel: Good point, I will do that in a separate PR

src/components/NameValueList/NameValueList.js
fspoettel: In this case, it would make sense to limit to `dt,dd` since we are inside a `dl`. Will add that

src/components/Tile/Tile.js
```diff
@@ -0,0 +1,106 @@
...
+  figure: null,
+  action: null,
+  titleText: '',
+};
+
+Tile.propTypes = {
+  /**
+   * HTML element name
+   */
+  as: PropTypes.string,
```
quadreex: not sure if we should impose limits here. We have a lot of places where we allow the string. Maybe we should go through all of the cases in `react-lds` in future and check where we should impose

fspoettel: I just checked, don't think we have a problem here:
* `Card` (default `article`) most block-level elements would make sense here depending on content
* `Column,Grid` even more so, any block-level tag should be allowed and makes sense. For `Column`, inline tags are also ok
* `Container` -> any block level element

PR #461, opened 2019-11-25
Add custom adjustments

PeterSiegl:


src/components/Chat/Chat.js
```diff
@@ -84,6 +84,7 @@ Chat.propTypes = {
       title: PropTypes.string.isRequired,
     }),
     avatar: PropTypes.string,
+    errorMessage: PropTypes.node,
```
quadreex: why do we need this on top-level chat component?

PeterSiegl: It's just for getting an overview of the props in the Storybook. It should be `error` btw.

src/components/Chat/Content/MessageMeta.js
```diff
@@ -49,20 +53,22 @@ export const MessageMeta = ({
     );
   }
 
-  return !isConsecutive && meta;
+  return !isConsecutive && metaCmp;
 };
 
 MessageMeta.propTypes = {
+  timestamp: null,
```
quadreex: MessageMeta.propTypes here should be MessageMeta.defaultProps

PR #464, opened 2019-11-29
Deprecate Button flavor "stretch", MediaObject prop "customTag"

### LINKED ISSUE #418: "stretch" is not a button flavor
See: https://www.lightningdesignsystem.com/components/buttons/#Sizing

It should be applicable in addition to - for example - a `brand` flavor, maybe using a `stretch` prop.

fspoettel:
Closes #418

### DIFF
--- a/src/components/MediaObject/MediaObject.js
+++ b/src/components/MediaObject/MediaObject.js
@@ -8,6 +8,7 @@ const renderFigure = (figure, classes) => (
 
 const MediaObject = (props) => {
   const {
+    as,
     bodyClassName,
     center,
     children,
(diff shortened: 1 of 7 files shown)


src/components/MediaObject/MediaObject.js
```diff
@@ -69,6 +71,10 @@ MediaObject.defaultProps = {
 };
 
 MediaObject.propTypes = {
+  /**
+   * Renders a customTag instead of a div
```
quadreex: maybe `customTag` better to right as `custom tag` because it is not a prop name anymore

PR #470, opened 2019-12-04
Add ability to disable resend button in Chat component

quadreex:


src/components/Chat/Chat.js
```diff
@@ -69,12 +71,14 @@ const Chat = ({
...
   className: null,
   isPastChat: false,
+  isResendDisabled: false,
   translations: DEFAULT_TRANSLATIONS,
 };
 
 Chat.propTypes = {
   className: PropTypes.string,
   isPastChat: PropTypes.bool,
+  isResendDisabled: PropTypes.bool,
```
PeterSiegl: Hast `isResendDisabled` to be a global prop or would it make sense to move it to the item level like `onResend`?

PR #474, opened 2019-12-10
Add support for null children in Tabs component

quadreex:


src/components/Tabs/ControlledTabs.js
```diff
@@ -44,14 +44,14 @@ class ControlledTabs extends PureComponent {
     styled: false,
   }
 
-  static getPanelState(children) {
+  static getChildrenArray = children => React.Children.toArray(children).filter(child => !!child);
```
fspoettel: Shouldn't `c => c` be sufficient here? `.filter` will cast to a boolean anyway

quadreex: It should, I just like when the code has all the conversions explicit. Easier to read

PR #475, opened 2019-12-10
Allow passing renderIcon fn to ScopedNotification

fspoettel:


src/components/ScopedNotification/ScopedNotification.js
```diff
@@ -4,10 +4,21 @@ import cx from 'classnames';
 import { Icon } from '../Icon';
 import { MediaObject } from '../MediaObject';
 
+const defaultIconRenderer = (theme, size) => (
```
quadreex: Maybe better to pass params ar an object? `{ theme, size }`
This way You can just pass function components to `renderIcon`

fspoettel: Yes good point, will change

src/components/ScopedNotification/ScopedNotification.js
```diff
@@ -23,15 +34,7 @@ const ScopedNotification = (props) => {
...
-      figureLeft={(
-        <Icon
-          icon="info"
-          sprite="utility"
-          size="small"
-          svgClassName={theme === 'dark' ? null : 'slds-icon-text-default'}
-          title="information"
-        />
-      )}
+      figureLeft={renderIcon({ theme, size: 'small' })}
```
quadreex: Still not an object in `defaultIconRenderer`. So it will break without passing this parameter

PR #478, opened 2019-12-19
Implement WelcomeMat

fspoettel:


src/components/WelcomeMat/WelcomeMat.js
```diff
@@ -0,0 +1,165 @@
...
+  renderDismiss: PropTypes.func,
+  /**
+   * Should render a `WelcomeMatProgress` element. Called with `({ steps })`
+   */
+  renderProgress: PropTypes.func,
+  /**
+   * Steps. If set to `false` the "Splash" variant will be rendered
+   */
+  steps: PropTypes.oneOfType([
+    PropTypes.oneOf([false]),
```
quadreex: That's what I don't like most about JS. Passing absolutely different things to one component property or to one function argument.
I'd prefer to have splash variant in case steps are not set.

src/components/WelcomeMat/helpers.js
```diff
@@ -0,0 +1,13 @@
+import PropTypes from 'prop-types';
+
+export const stepType = PropTypes.shape({
```
quadreex: this is not a function. why is it in helpers and not in constants?

src/components/WelcomeMat/WelcomeMat.js
fspoettel: Hm it depends, overloading a prop/argument can sometimes make sense. I think in this case, overloading steps with `false` is clearer than `undefined`. `steps={[]}` and `steps={undefined}` should render the same to avoid bad surprises. Setting `steps=false` is more expressive than passing `steps={/*ignored&*/} isSplash` imo, but that would be another option that we could think about.

quadreex: IMHO not giving `steps` prop to the component at all is more clear way to say that we want splash. And `undefined` and empty array are also different things, and so should be treated and rendered differently.
Also `steps={false}` to me looks like:
1) violation of DOM standard prop passing.
2) violation of semantics. Wording "steps" means we expect a collection of something. And `false` is not a collection.

src/components/WelcomeMat/components/WelcomeMatContent.js
```diff
@@ -0,0 +1,155 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import { stepType } from '../constants';
+import { Column, Grid } from '../../Grid';
+import { WelcomeMatStep } from './WelcomeMatStep';
+
+class WelcomeMatContent extends Component {
```
quadreex: this component can be done as a functional component. Can we use hooks in `react-lds`?

fspoettel: Eventually yes, it would require a major version bump though as going from requiring `react@.6` to `react@.8` is a breaking change. We should def. do it in the near future though imo

PR #480, opened 2020-01-17
Fix layout of setup assistant

PeterSiegl:
https://www.lightningdesignsystem.com/components/setup-assistant/

### DIFF
--- a/src/components/SummaryDetail/SummaryDetail.js
+++ b/src/components/SummaryDetail/SummaryDetail.js
@@ -10,32 +10,36 @@ const SummaryDetail = ({
   onOpen,
   title,
   renderTitle,
-}) => (
-  <div className={cx('slds-summary-detail', { 'slds-is-open': isOpen })}>
-    {onOpen && (
-      <IconButton
-        onClick={onOpen}
-        className={cx('slds-m-right_x-small', iconButtonClassName)}
-      >
-        <ButtonIcon
-          className="slds-summary-detail__action-icon"
-          icon="switch"
-          sprite="utility"
-        />
-      </IconButton>
-    )}
-    <div>
-      <div className="slds-summary-detail__title">
-        {renderTitle(title)}
-      </div>
-      {children && (
-        <div aria-hidden={!isOpen} className="slds-summary-detail__content">
-          {children({ isOpen })}
-        </div>
+}) => {
+  if (!onOpen) return renderTitle(title);
+
+  return (
+    <div className={cx('slds-summary-detail', { 'slds-is-open': isOpen })}>
+      {onOpen && (
+        <IconButton
+          onClick={onOpen}
+          className={cx('slds-m-right_x-small', iconButtonClassName)}
+        >
+          <ButtonIcon
+            className="slds-summary-detail__action-icon"
+            icon="switch"
+            sprite="utility"
+          />
+        </IconButton>
       )}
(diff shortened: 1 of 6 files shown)


src/components/SummaryDetail/SummaryDetail.js
```diff
@@ -10,32 +10,36 @@ const SummaryDetail = ({
...
-    <div>
-      <div className="slds-summary-detail__title">
-        {renderTitle(title)}
-      </div>
-      {children && (
-        <div aria-hidden={!isOpen} className="slds-summary-detail__content">
-          {children({ isOpen })}
-        </div>
+}) => {
+  if (!onOpen) return renderTitle(title);
```
fspoettel: This seems to directly contradict the example on https://lightningdesignsystem.com/components/summary-detail/#Closed

src/components/SummaryDetail/SummaryDetail.js
```diff
@@ -10,32 +10,36 @@ const SummaryDetail = ({
...
+          className={cx('slds-m-right_x-small', iconButtonClassName)}
+        >
+          <ButtonIcon
+            className="slds-summary-detail__action-icon"
+            icon="switch"
+            sprite="utility"
+          />
+        </IconButton>
       )}
+      <div className="slds-container_fluid">
```
fspoettel: I can't find this anywhere in the documentation on https://lightningdesignsystem.com/components/summary-detail. Where does this come from?

PeterSiegl: Screenshot 2020-01-17 at 11 16 28

src/components/SummaryDetail/SummaryDetail.js
PeterSiegl: In the specifications the `Base` and the `With Step Progress` layout of the `SetupAssistant` do not have the `slds-summary-detail` wrapper. It's only needed for the `Hub with Expandable Steps`.

PeterSiegl: The initial problem that lead me to this fix was that the actions on the right were left aligned if there was less text. Following the specifications the actions should be right aligned.
Screenshot 2020-01-17 at 11 27 53

fspoettel: That all makes sense in the case of `SummaryDetail` being used in a `SetupAssistant`. However, the documentation for SummaryDetail outlines different markup for both the base component and the closed state. Since `SummaryDetail` is a standalone component, that standalone component now uses different markup than documented. In my eyes, this fix only applies to `SetupAssistant` while potentially breaking the standalone component.

PeterSiegl: Makes sense, I will solve this issue within the `SetupAssistant` component.

src/components/SummaryDetail/SummaryDetail.js
```diff
@@ -24,7 +24,7 @@ const SummaryDetail = ({
         />
       </IconButton>
     )}
-    <div>
+    <div className="slds-container_fluid">
```
fspoettel: This class is not present in the standalone `SummaryDetail`. We could add a `containerClassName` prop or similar to set it in `SetupAssistant`.

PeterSiegl: An additional `containerClassName` prop sounds good! 👍

PR #488, opened 2020-04-10
Fix tab links missing key attributes

### LINKED ISSUE #484: Tabs: it's not possible to pass a dynamic set of children without getting the "unique key prop" error
That's because the passed `Tab` components are mapped to something else and the key prop is omitted

fspoettel:
Closes #484

I adjusted the story so this can be tested for properly.


fspoettel:
Thanks for the review @philippfrank. I will prepare a patch release `v9.1.1` later today

